PR #19, opened 2020-07-22
Add options to send webp paths to lottie-scroll directive

jeremydw:
This is a stopgap until we have deeper Google image service support for Lottie files. This allows the user to manually specify a JSON path to the Lottie.

Is this OK to add here? An alternative design would be to replace the JSON path within the Lottie object beforehand with the WebP path, but I think I like this better because it keeps the data structure in YAML and the JS implementation coupled.


uxder:
Thank you! Going out with v216.

PR #22, opened 2020-08-04
Fix race condition with disposed blob loaders

stevenle:


uxder:
Thank you

uxder:
Available in v239.

PR #35, opened 2020-11-27
Add UidIterator

angusm:


uxder:
Thanks, looks pretty straight forward but can you add just a simple example?

PR #38, opened 2020-11-27
Add arrayf.flatten function

angusm:


uxder:
LGTM.  Looks like there are merge issues so I'll leave it up to you to resolve.  Thank you so much!

PR #42, opened 2020-11-27
Add mathf.getSign

angusm:
If yano-js does not need to support IE we can ignore this
and use Math.sign directly. Not sure what support we want
to consider.

### DIFF
--- a/src/mathf/mathf.ts
+++ b/src/mathf/mathf.ts
@@ -1397,4 +1397,10 @@ export class mathf {
     return mathf.lerp(minA, maxA, t);
   }
 
+  static getSign(x: number): number {
+    return !Math.sign ?
+        // Use MDN polyfill for math.sign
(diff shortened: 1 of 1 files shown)


jeremydw:
I think we can abandon IE support; but this raises a good question – we should establish a browser compatibility matrix. I'm thinking we want to support only Evergreen browsers minus two major versions, so:

- Chrome
- Safari
- Edge (Chromium version)
- Firefox

@uxder @stevenle – thoughts?

If we agree on the above I'll doc it.

uxder:
The above browser matrix sounds good to me 👍

jeremydw:
I've added that to the README:  
https://github.com/grow/yano-js/commit/7b440f9c0595ffc436ba9ee574b4a0cbb79153bc

angusm:
Abandoning per discussion. Preference is to use Math.sign.

PR #47, opened 2020-11-27
Add iterableIteratorf and functions

angusm:


uxder:
Do you think you can add some examples at the top of class?

PR #48, opened 2020-11-29
Add support for Color names

angusm:
This might be a call on weight, whether we want to have this functionality for users or insist on HEX/RGB values instead.


uxder:
Thanks Angus for adding this.

@jeremydw @stevenle Do you have an opinion on this?  I personally never use color names so am thinking we perhaps omit it but we can include it if others find useful.

stevenle:
i personally would prefer to keep the library small (in size and in memory usage) and the map looks fairly big for functionality that could be moved to where it's being called.

jeremydw:
How about extending Yano just enough so that the colors can be defined
locally within the project where this functionality is being used? Then, if
it needs to be used in more than N projects (let's use 2 for N?) we can
move the functionality into the core library.

@All, how does that sound as a method of governing what makes it into the
core library? Are folks happy with "2"?

PR #49, opened 2020-11-29
Add Map classes

angusm:


uxder:
Look awesome 👍     I think we can merge but do you think we should create some tests for this one?

PR #54, opened 2020-12-15
Add Dimensions2d class

angusm:


src/mathf/dimensions-2d.ts
```diff
@@ -0,0 +1,60 @@
+import { Vector } from './vector';
+import { SCROLL_ELEMENT } from '../dom/scroll-element';
+import { ROOT_ELEMENT } from '../dom/root-element';
+
+export class Dimensions2d extends Vector {
```
uxder, line 5: This is tough one.

So in my mind, mathf tends to be pure math functions.  But this class is technically, DOM related so I find it more suitable under /dom and perhaps something like DomVector something.  WDYT?

src/mathf/dimensions-2d.ts
```diff
@@ -0,0 +1,60 @@
...
+  static fromRootElement<T extends Dimensions2d>() {
+    return <T>new this(ROOT_ELEMENT.clientWidth, ROOT_ELEMENT.clientHeight);
+  }
+
+  static fromScrollElementClient<T extends Dimensions2d>() {
+    return <T>new this(SCROLL_ELEMENT.clientWidth, SCROLL_ELEMENT.clientHeight);
+  }
+
+  static fromInnerWindow<T extends Dimensions2d>() {
+    return <T>new this(window.innerWidth, window.innerHeight);
```
uxder, line 45: I think we need to test for window - if we want to be able to use yano in a pure node environment (where window maybe undefined).

src/dom/root-element.ts
```diff
@@ -0,0 +1,2 @@
+export const ROOT_ELEMENT = document.children[0];
```
uxder, line 1: What's the role of this guy?

uxder: And also the scroll-element?  Is this just pure caching of the reference?

angusm: Yes, purely caching. Moved SCROLL_ELEMENT to a function, ditched ROOT_ELEMENT.

src/user-agent/browser/edge.ts
```diff
@@ -0,0 +1,11 @@
+import {Browser, TOffset} from "./base";
```
uxder, line 1: Have you checked out the yano `is` class?  A majority of these user-agent tests are covered there....is that something that would work for you?

src/user-agent/feature-support/is-webp-supported.ts
```diff
@@ -0,0 +1,23 @@
...
+  typeof Browser,
+  NumericRange
+>([
+  [Edge, new NumericRange(18, Number.POSITIVE_INFINITY)],
+  [Firefox, new NumericRange(65, Number.POSITIVE_INFINITY)],
+  [Chrome, new NumericRange(32, Number.POSITIVE_INFINITY)],
+  [Opera, new NumericRange(19, Number.POSITIVE_INFINITY)]
+]);
+
+function isWebpSupported(browser: typeof Browser): boolean {
```
uxder, line 19: Kind of a duplicate of is.supportigWebp() and is.supportingWebpAsync().

Although this version might be faster since just checks for browser version number.

uxder (approved):
@angusm it looks good.

I see some duplication with the yano IS class.  So check that out if you haven't yet.

I'm okay with adding the user-agent stuff since it in a subdirectory /user-agent and might be good for cases where you just want pieces.

Please check this since you need it but could you later go back and add some documentation for some of the clases?  Just a comment at the top of each class and some quick examples helps us read through it.

src/mathf/dimensions-2d.ts
angusm: Created DOM variants.

src/mathf/dimensions-2d.ts
angusm: Moved window functions to the DOM variant. Assuming DOM functions aren't used in Node? Apologies for my ignorance, I've done very little pure node.

src/user-agent/browser/edge.ts
angusm: Yes, removed user-agent.

src/user-agent/feature-support/is-webp-supported.ts
angusm: Removed user-agent/

angusm:
Went through with aims to consolidate and comment.

src/dom/computed-style-service.ts
```diff
@@ -0,0 +1,42 @@
...
+    return this.computedStyle.get(element);
+  }
+
+  private init() {
+    this.raf.start();
+  }
+
+  private loop() {
+    this.raf.postWrite(() => this.computedStyle.clear());
+  }
```
uxder, line 41: I think we need to to dispose of raf?

src/arrayf/arrayf.ts
```diff
@@ -68,4 +68,45 @@ export class arrayf {
         return maxValue;
     }
 
+    /**
+     * Determines if the value in each index of each list matches the
+     * corresponding value in that same index in each other list.
+     * @param lists
+     */
+    static areArrayValuesIdentical<T>(lists: T[][]): boolean {
```
uxder, line 76: I'm not sure of the performance but I wonder if is the same thing and faster?

return JSON.stringify(a)==JSON.stringify(b);

src/dom/cached-element-vector.ts
```diff
@@ -0,0 +1,157 @@
+/**
+ * CachedElementVector is used to optimize access to commonly checked values.
```
uxder, line 2: Nice.  Thanks for adding comments.  Make its really clear and I think want to use this.

src/dom/cached-element-vector.ts
```diff
@@ -0,0 +1,157 @@
...
+        1000);
+  }
+
+  // Expected to be overridden
+  protected getValues(): number[] {
+    throw new Error('getValues must be overridden by child class');
+  }
+
+  private init(): void {
+    this.raf.start();
```
uxder, line 131: Make sure we dispose raf.

src/dom/get-scroll-element.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Handle variations between browsers that are relevant when measuring the
+ * position of elements relative to one another.
+ */
+export function getScrollElement(): Element {
```
uxder, line 5: Could this move to dom?  I think we were trying to reduce file count?

src/dom/position/matrix.ts
```diff
@@ -0,0 +1,182 @@
...
+const preChangeMatrixByElement: Map<HTMLElement, Matrix> = new Map();
+
+// For figuring out how much to trim off strings to get at the values
+const STYLE_STRING_PREFIX = 'matrix(';
+const STYLE_STRING_PREFIX_LENGTH = STYLE_STRING_PREFIX.length;
+
+// Optimization to minimize DOM queries where possible
+const computedStyleService = ComputedStyleService.getSingleton();
+
+export class Matrix {
```
uxder, line 24: My gut tells me maybe we should call this DomMatrix since it's specialized to that purpose.

uxder (commented):
Looks like x/x.ts was deleted.

uxder:
@angusm Nice work.  I think some of these classes are useful.

One overall comment, I have is that a lot of the caching classes might come at the cost of too much abstraction making it harder for others to navigate through code without heavily investing learning the library.

Part of me says, we should keep it simple and use straight native javascript read / writes combined in raf.read / raf.write - which isn't as optimized but simpler and easier to follow.    However, I do see a real need to cache things like scroll, window so I see those as very useful.

I think this is good to go but since it does come at a cost of abstraction, just want to make sure we continue to make efforts to add docs / examples / tests to ease the cost of onboarding for others.

Thank you!

src/arrayf/arrayf.ts
angusm: It looks like JSON.stringify is very slow by comparison: https://jsbench.me/z5kitbq0tr/1

src/dom/cached-element-vector.ts
angusm: It gets disposed on line 118 in destroy()

But that reminded me of another thing I meant to do.

I renamed `destroy()` to `dispose()` to align w/ yano convention.

src/dom/computed-style-service.ts
angusm: Resolved. Made it track its uses and dispose itself after all of its uses have called dispose and its been idle for a bit.

src/dom/get-scroll-element.ts
angusm: Done.

src/dom/position/matrix.ts
angusm: Done.

Used MatrixDom to following existing convention from VectorDom

uxder:
> @angusm Nice work. I think some of these classes are useful.
> 
> One overall comment, I have is that a lot of the caching classes might come at the cost of too much abstraction making it harder for others to navigate through code without heavily investing learning the library.
> 
> Part of me says, we should keep it simple and use straight native javascript read / writes combined in raf.read / raf.write - which isn't as optimized but simpler and easier to follow. However, I do see a real need to cache things like scroll, window so I see those as very useful.
> 
> I think this is good to go but since it does come at a cost of abstraction, just want to make sure we continue to make efforts to add docs / examples / tests to ease the cost of onboarding for others.
> 
> Thank you!

@jeremydw @stevenle  - have any thoughts on this?

angusm:
Would adding helper functions to dom like there is currently for `getStyle` rather than using the singletons directly help make this easier to work with?

Sort of bury the caching in a black box users don't need to worry about, beyond "use these functions instead of native calls for best results"?

src/types.ts
```diff
@@ -0,0 +1 @@
+export type Numeric = number|string;
```
stevenle, line 1: i don't think you should have this in a generic location this unless it explicitly fills the contract in a generic yet explicit way...can you remove this and move this into the file where it's being used (or just use `number | string` directly without defining a custom type)?

src/iterable-iterator/map.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Maps the values in the iterator to an array using the given function.
+ * Behaves similarly to Array.map.
+ */
+export function map<T, V>(
```
stevenle, line 5: all of these iterable utility functions don't need to be in their own files, should be consolidated into a single file. @uxder maybe can suggest a naming for it, but if we're following the current structure, maybe something like:

src/iterables/iterables.ts

then the import/usage can be something like:

```
import * as iterables from 'iterables';

iterables.map(...);
```

src/dom/webgl-image-sequence.ts
```diff
@@ -991,6 +990,12 @@ export class WebGlImageSequence {
         if (imageSource == this.lastDrawSource) {
             return;
         }
+
+        // Can't draw on a 0 area canvas
+        if (this.canvasWidth * this.canvasHeight === 0) {
```
stevenle, line 995: small detail but instead of multiplying the values (which can create a large number and would potentially need to create more memory for it), can you simply check each value instead? e.g.:

```
if (this.canvasWidth === 0 || this.canvasHeight === 0) { ... }
```

src/mathf/geometry/multi-dimensional-vector.ts
```diff
@@ -0,0 +1,191 @@
+import { arrayf } from '../../arrayf/arrayf';
+import { mathf } from '../mathf';
+
+/**
+ * Class for handling vectors of arbitrary dimensions
+ */
+export class MultiDimensionalVector {
```
stevenle, line 7: just a general note: it's really hard for me to follow this code mainly because everything is so heavily abstracted and has generic typings, etc.. i would suggest maybe looking at similar code and seeing if this code can be simplified.

for example, instead of accepting generic types, is it at all appropriate to accept explicit types?

or if `Dimensions2d` is the only implementation of `MultiDimensionalVector` right now, i'd suggest consolidating everything into the 2d class until we have a need for this base class.

compare to:

https://github.com/maxkueng/victor/blob/master/index.js

which is a little easier for me to read and follow.

src/arrayf/arrayf.ts
```diff
@@ -68,4 +68,45 @@ export class arrayf {
...
+        return values.length === 0 ||
+            values.every((value) => values[0] === value);
+    }
+
+    /**
+     * Removes the first instance of a value from the given array.
+     * Useful when tracking the places a singleton or service is used so that
+     * it can dispose of itself when no longer needed.
+     */
+    static removeFirstInstance<T>(values: T[], value: T): T[] {
```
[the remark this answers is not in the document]
angusm: The behaviour is different.

I will add a better example to the code comment.

src/types.ts
angusm: Done.

src/iterable-iterator/map.ts
angusm: Can do.

jeremydw:
Just to make sure I'm chiming in and adding my bits. This is general feedback I want us all to apply to get on the same page with our code style and approach. :)

- Generally, create fewer files, with code grouped logically into the same file works best. Fewer files should hopefully reduce cognitive overhead when trying to determine where to go to update something, or learn how something works.
- Avoid reimplementing things that are available from the standard library.
- Avoid too much inheritance. I think inheritance can be really clever in certain cases but often for contributors it will require someone to have to trace through many files to read the code end to end. If a concept can be expressed with a standard "or" statement (picking on `DynamicDefaultMap` here) then I think the "or" statement is preferred. For `DynamicDefaultMap` it's pretty hard for me to follow.
- I recently wrote a Contributing guide for Amagaki that I think can apply here to Degu (new name) too. I'll work to send a PR that adapts the Contributing guide for Degu, but I think it has some good learnings: https://github.com/blinkkcode/amagaki/blob/main/CONTRIBUTING.md#philosophy -- my favorite takeaway is "Write as little code as possible, but as much as necessary to complete a task correctly."

src/arrayf/arrayf.ts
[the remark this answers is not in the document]
stevenle: ah okay, sorry i misread the documentation. one optimization i'd suggest is removing the element in place so you don't need to create a new array and copy over elements from the old one.

i checked closure's implementation of the same and they do:
https://github.com/google/closure-library/blob/master/closure/goog/array/array.js#L695

```
/**
 * Removes the first occurrence of a particular value from an array.
 * @param {IArrayLike<T>} arr Array from which to remove
 *     value.
 * @param {T} obj Object to remove.
 * @return {boolean} True if an element was removed.
 * @template T
 */
function remove(arr, obj) {
  var i = indexOf(arr, obj);
  var rv;
  if ((rv = i >= 0)) {
    removeAt(arr, i);
  }
  return rv;
}
exports.remove = remove;


/**
 * Removes from an array the element at index i
 * @param {IArrayLike<?>} arr Array or array like object from which to
 *     remove value.
 * @param {number} i The index to remove.
 * @return {boolean} True if an element was removed.
 */
function removeAt(arr, i) {
  asserts.assert(arr.length != null);

  // use generic form of splice
  // splice returns the removed items and if successful the length of that
  // will be 1
  return Array.prototype.splice.call(arr, i, 1).length == 1;
}
```

src/dom/webgl-image-sequence.ts
angusm: Nice. Changed to` if (!this.canvasWidth || !this.canvasHeight) {`

src/mathf/geometry/multi-dimensional-vector.ts
angusm: In my head it was a trade-off between ease-of-use working within the abstract class' code, but gaining ease-of-use working with the instances of its extended classes and narrowing maintenance of generic behaviour to a single point.

By being this generic it:
- Supports 2d/3d vectors, areas, volumes and colors
- Limits maintenance of common functions to a single point
- Limits testing of common functions to a single point
- Provides a consistent API and behaviour across vectors

3d vectors and Colors are still sitting in Toolbox waiting to be ported or consolidated with the existing Yano classes, since their need wasn't as urgent for current work.

The cons are:
- Readability of generic/abstracts
- Full benefit would depend on eventually having the existing Vector/Color classes in Yano extend this base

If this is acceptable we can leave it as is.

We can simplify a bit if `: MultiDimensionalVector` is used instead of `: <T extends MultiDimensionalVector>`, casting returned values to the desired type.

We can remove this class and duplicate functionality as needed into child classes, accepting potentially different behaviour per vector. API could be kept consistent via an Interface.

We could use an Interface to keep the API consistent and remove the generic class, replacing it with a decorator that adds the duplicated functionality to classes as needed.

We could use an interface to keep the API consistent, remove the generic class, treat instances as immutable data types and use static functions on classes for manipulations. Essentially having a vectorf type class.

So we can:
- Leave as is, add a note that this is not intended to be used directly
- Replaces instances of `T extends MultiDimensionalVector` with `MultiDimensionalVector` and cast extended types as needed.
- Remove abstract class, duplicate code into explicit classes
- Remove abstract class, create decorator
- Remove abstract class, treat instances as immutable data types, use static functions for generic behaviour.

angusm:
Let me re-work DynamicDefaultMap. It's meant to behave as defaultdict from python.
https://www.accelebrate.com/blog/using-defaultdict-python#:~:text=A%20defaultdict%20works%20exactly%20like,returned%20by%20the%20default%20factory.

The inheritance for these map classes was such a mess because at one point ES6 Map wouldn't allow itself to be extended directly. https://github.com/Microsoft/TypeScript/issues/10853 That seems to have been fixed in recent browser updates :) so this can be made much less messy.

Can you be specific about `Avoid reimplementing things that are available from the standard library.`?

I agree fully with `Write as little code as possible, but as much as necessary to complete a task correctly.`, all of this code has been necessary at some point, it was left in to try and be pro-active against future needs, but I will go through and remove any pieces not currently in active use or needed immediately.

src/iterable-iterator/map.ts
uxder: Thanks Angus - do you think we can actually make that change against some of the recent stuff you checked in as well (can be a separate PR to avoid overloading this one).  

map-wrapper, mapped-iterator, maulti-value-array etc per trying to lower file count and keeping it all contained.

src/arrayf/arrayf.ts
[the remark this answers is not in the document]
angusm: Simplified and moved to MutliDimensionalVector

src/arrayf/arrayf.ts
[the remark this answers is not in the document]
angusm: Gotcha. I was trying to be more functional and immutable with stuff at the time I wrote this.

Looking at it now, nothing depends on those properties and its currently only used one place so I am going to move this to the relevant file and out of arrayf and make that change.

src/iterable-iterator/map.ts
angusm: Created issue for myself to keep current PR on track: https://github.com/grow/yano-js/issues/55

src/dom/cached-element-vector.ts
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * All instances of CachedElementVector will share a single cache to facilitate
+ * abstraction of the cache checking/clearing.
+ *
+ * Also helps ensure that only singletons are being created by checking for
+ * existing instances in this cache.
+ */
+const caches: DynamicDefaultMap<any, InnerCache> =
+    DynamicDefaultMap.usingFunction<any, any>(
```
stevenle, line 24: this code is pretty hard for me to read, maybe because there's so much generics here, and maybe the api for the "default map" could be improved.

i wonder if you can move the generics handling to where the map values are being fetched instead, which would simplify things a little and make it generally easier to read. or maybe use a regular constructor so that the generics type setting isn't happening so much in the function call.

something like:

```javascript
class DefaultMap extends Map {
  getOrSet(key: string, value: any | () => any): any { ... }
}

const map = new DefaultMap();
const value = map.getOrSet('key', () => { return someDefaultValue }) as SomeType;
```

i think with something like this, i'd be easier for me to understand, like "ok this is a map, and if a value doesn't exist, it sets it."

with the example above i used `any` but generics could/should be used, i just didn't know the syntax off the top of my head. i'm guessing the user call would be something like:

```javascript
const map: DefaultMap<string, SomeType> = new DefaultMap();
const value = map.getOrSet('key', () => { return someDefaultValue });  // returns a `SomeType`
```

src/mathf/geometry/multi-dimensional-vector.ts
stevenle: i think for me personally i would love for this module to be simplified somehow. as someone not super familiar with this code, i would have a hard time figuring out how to instantiate a vector and to call its appropriate functions.

i wish i had better suggestions, but i haven't used vectors too much. mind checking closure library and the aforementioned `victor` lib to see if there are potentially better ways to structure this?

https://github.com/google/closure-library/tree/master/closure/goog

if you're pressed for time and need to get this in, please don't let me block this review.

uxder: Angus - in light of trying to reduce abstraction and keeping it simple, can you use yano.vector instead?  

IMO, I think the mix of static and non static classes tends to be an overhead and it's just simpler to keep that straight-forward instead of doing what has been done here: https://github.com/grow/yano-js/blob/821fea82a320f1b13a0c4c3e205a116ab3a36aa9/src/mathf/geometry/multi-dimensional-vector.ts#L75 but perhaps there was a reason for this.

How often do you use Dimensions2D?  it seems to basically be just a vector2D except calling x and y  - width and height instead so I didn't really see a strong need for that but correct me if I'm wrong.

angusm: The closure library does what I was hoping to avoid, which is duplicating code and creating multiple maintenance and test points.

It's not intended to be used directly, but as a base for Vector2d Dimensions2d etc.

It would be used like:

```
new MultiDimensionalVector(1, 0, 7).add(new MultiDimensionalVector(2, 3, -1));

// Returning equivalent of new MultiDimensionalVector(3, 3, 6)
```

But given more explicit values seems to be the preferred method I'll remove this class and re-create its functionality in its sub-classes as needed.

src/dom/computed-style-service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  dispose(use: any) {
+    const useIndex = this.uses.indexOf(use);
+    if (useIndex === -1) {
+      return;
+    }
+    this.uses.splice(useIndex, 1);
+    // Hang tight for a minute so we aren't creating and destroying the
+    // singleton too rapidly.
```
uxder, line 60: I didn't follow why we need this as opposed to just running it.

src/dom/cached-element-vector.ts
angusm: This would go against the intent of the default map which is to avoid duplicated conditional get calls by assuming them as the default.

I'll remove this base class and re-implement its functions in its child classes as needed since this follows the same pattern as MultiDimensionalVector does.

src/dom/position/dimensions.ts
```diff
@@ -0,0 +1,33 @@
+import { CachedElementVector } from '../cached-element-vector';
+import { Dimensions2dDom } from './dimensions-2d-dom';
+
+/**
+ * Caches element dimensions on a per-frame basis.
+ * Useful for measuring element distances as high level elements may be queried
+ * multiple times.
+ */
+export class Dimensions extends CachedElementVector<Dimensions2dDom> {
```
uxder, line 9: So after further thought, I'm kind of more in the opinion that perhaps we don't need this level of caching in the library itself.  Sorry for going back and forth on this.

I think we want to uphold performance but trying to manage and caching at a PER frame level adds a lot of complexity and abstraction.  I think we want to keep things fairly simple here.

For sties and cases that may require this level of per frame cache optimization, these services should be within the project and not part of this library.

WTDY?

uxder (commented):
Added a few more comments.

src/dom/cached-element-vector.ts
stevenle: i could be wrong, but i think if you wrap the call in a fn, it won't get evaluated until the callback is called.

src/dom/position/dimensions.ts
stevenle: i wonder if instead of having separate caches for each thing that may need caching, if we can instead have a single cache.

something like:

```javascript
const cache = new RafCache();  // gets reset on every frame
const windowHeight = cache.get('window.innerHeight', () => window.innerHeight);
```

then there wouldn't need to be any code to track "usages" of different types of values, and this code could be moved to the caller instead of having it within this library itself.

stevenle: i've been giving it more and more thought and i tend to agree with scott that maybe the cache-related code here may not make sense. it creates a level of interconnectedness across different parts of the code that, while may optimize things, makes the individual utility functions more difficult to maintain.

for example, with the change to `isDisplayNone()`, the function no longer simply tells you whether an element is set to `display:none`. if i understand the code correctly, it now fetches the style service (instantiates it if it isn't already instantiated), allocates a cache object, registers the cache object in its "usages" map, and subscribes to raf to clear its values on every frame.

for the most part, i think the library should focus on providing utilities that have simple, clear, easy-to-use apis. the extra classes that hyper optimize the dom are still useful, but maybe they should be moved into the project directories themselves, which would make it easier to maintain since those apis can be changed out rather easily without affecting any other clients who may also be using it.

src/dom/cached-element-vector.ts
angusm: It wasn't a performance concern as much as it was a developer ergonomics concern. I found I was doing a lot of boilerplate, particularly with caching, that could be eliminated with this class.

Part of the reason for writing this class was to avoid repeating `map.get(foo) || (foo) => someThingWithFoo(foo)`, or having to add private methods to each class to the same effect.

src/dom/position/dimensions.ts
angusm: I will yank them. The uses/disposal could probably be done neater. I'm going to leave the benchmark here so we have a rough idea of potential gains if we ever need to revisit performance out of necessity: https://jsbench.me/eokiv6f8qy/1

src/dom/dimensions-2d-dom.ts
```diff
@@ -0,0 +1,21 @@
...
+    this.height = height;
+  }
+
+  equals(candidate: Dimensions2dDom) {
+    return this.width === candidate.width &&
+        this.height === candidate.height;
+  }
+
+  getHypotenuseLength() {
+    return Math.sqrt(Math.pow(this.width, 2) + Math.pow(this.height, 2));
```
stevenle, line 19: `this.width * this.width` will be significantly faster than `Math.pow(this.width, 2)`

https://www.measurethat.net/Benchmarks/Show/1651/0/math-pow-vs-multiply

src/map/default.ts
```diff
@@ -0,0 +1,37 @@
...
+              defaultFunction: DefaultFunction<K, V> = noop) {
+    super(iterable);
+    this.defaultFunction = defaultFunction;
+  }
+
+  get(key: K): V {
+    if (!this.has(key)) {
+      this.set(key, this.defaultFunction(key));
+    }
+    return super.get(key);
```
stevenle, line 35: Although a `map.get()` is generally expected to be `O(1)`, there's a worst-case scenario of it being `O(N)` depending on its implementation. Since you deterministically know the value above in the `this.set()` line, that value should be used instead of doing a second lookup.

src/dom/get-visible-y-from-root.ts
```diff
@@ -0,0 +1,47 @@
...
+function getTransformedOffset(candidateElement: HTMLElement): number {
+  return candidateElement.offsetTop +
+      domVectorf.fromElementTransform(candidateElement).y;
+}
+
+function getVisibleDistanceFromRoot(
+    element: HTMLElement,
+    getOffsetFn: (e: HTMLElement) => number
+): number {
+  // Short circuit for fixed elements
```
stevenle, line 19: Small nit pick: Google style guides generally recommend comments that end in punctuation.

src/dom/get-visible-y-from-root.ts
```diff
@@ -0,0 +1,47 @@
+import { dom } from '../';
+import { domVectorf } from './dom-vectorf';
+
+const ignoredPositions = new Set(['fixed', 'absolute']);
```
stevenle, line 4: Where is this being used?

src/dom/dom.ts
```diff
@@ -849,16 +847,14 @@ export class dom {
      *
      * ```
      */
-    static getStyle(el: Element): CSSStyleDeclaration {
-        return el['currentStyle'] || window.getComputedStyle(el);
+    static getStyle(el: Element): CSSStyleDeclaration {return el['currentStyle'] || window.getComputedStyle(el);
```
stevenle, line 850: Revert this change?

src/dom/matrix-dom.ts
```diff
@@ -0,0 +1,55 @@
+/**
+ * Class for dealing with CSS Transform matrices.
+ * Supports relative updates being made to a matrix from multiple pieces of
+ * code without them over-riding each other (unless explicitly desired).
+ */
+import { dom } from './dom';
+
+type Numeric = number|string; // Number that could be represented as string
```
stevenle, line 8: I've been thinking about this some more, and since typescript is strict about types, I think it's better that only number values are accepted below and that users of this class (including `parseFromString`) are required to pass in the correct type.

Otherwise passing an invalid value like `'a'` will produce `NaN` and cause problems.

stevenle:
@uxder, i dunno what your file naming conventions are, but i'm thinking maybe `map/default.ts` should be renamed `map/default-map.ts` since the term "default" can have a confusing meaning here. wdyt?

src/dom/dimensions-2d-dom.ts
angusm: Nice. Done :)

src/map/default.ts
angusm: Nice catch :) Done

src/dom/get-visible-y-from-root.ts
angusm: Done.

src/dom/get-visible-y-from-root.ts
angusm: It isn't anymore, sorry for missing this.

There was code for getting what the position of a sticky element would be if it wasn't stickied. This wasn't an immediate need so I removed it from this PR but this variable squeaked through unnoticed.

Removed.

src/dom/dom.ts
angusm: Apologies, must've hit a wrong key there.

angusm:
> @uxder, i dunno what your file naming conventions are, but i'm thinking maybe `map/default.ts` should be renamed `map/default-map.ts` since the term "default" can have a confusing meaning here. wdyt?

Yes. Done.

src/dom/matrix-dom.ts
angusm: Done. Added a check in parseFromString to make sure the values grabbed are appropriate length and not-NaN, throwing an error if used with incorrect values.

src/dom/get-visible-y-from-root.ts
```diff
@@ -0,0 +1,45 @@
...
+function isFixed(element: HTMLElement) {
+  return dom.getStyle(element).position === 'fixed';
+}
+
+function getTransformedOffset(candidateElement: HTMLElement): number {
+  return candidateElement.offsetTop +
+      domVectorf.fromElementTransform(candidateElement).y;
+}
+
+function getVisibleDistanceFromRoot(
```
uxder, line 13: Can you add usage examples?  Also consider using words ancestor as I think Dom classes use that as  a common term.

uxder, line 13: Can you add usage examples?  Also consider using words ancestor as I think Dom classes use that as  a common term.

src/dom/dimensions-2d-dom.ts
```diff
@@ -0,0 +1,21 @@
+export class Dimensions2dDom {
```
uxder, line 1: I assume this is the start of the class?  As is I don't feel we should check this in as it doesn't add too much value but I can see is building more methods to this to make it a useful tool.

uxder (approved):

src/dom/get-visible-y-from-root.ts
angusm: Added examples. Renamed for clarity.

src/dom/dimensions-2d-dom.ts
angusm: I think the value is in readability w/o additional comments versus actual functionality.

```
  const dimensions = Dimensions2dDom.fromElementOffset(this.el);
  if (this.lastDimensions && dimensions.equals(this.lastDimensions)) {
    return;
  }
  ```

  Reads clearer to me than...
  
  ```
  const dimensions = [this.el.offsetWidth, this.el.offsetHeight];
  if (
      this.lastDimensions &&
      this.lastDimensions[0] === dimensions[0] &&
      this.lastDimensionsp[1] === dimensions[1]
  ) {
    return;
  }
  ```
  
  That said I can solve this with comments, so I will remove this for now.

PR #59, opened 2021-01-07
Update package name and auto-publish

jeremydw:
- Automatically publish to npmjs.org whenever the package version is incremented in `package.json` and committed to branch `main`
- Also updates package name to `@blinkk/degu` (eventually we will move blinkkcode on GH to blinkk)


uxder (approved):
LGTM 👍

PR #63, opened 2021-01-19
Add capture to scroll-render-fix.

uxder:
@angusm Can you take a look when you have a chance?  Thank you!

### DIFF
--- a/src/dom/scroll-render-fix.ts
+++ b/src/dom/scroll-render-fix.ts
@@ -55,7 +55,7 @@ export class ScrollRenderFix {
         this.domWatcher.add({
             element: window,
             on: 'wheel',
-            eventOptions: { passive: false },
+            eventOptions: { passive: false, capture: true },
             callback: this.wheelHandler.bind(this)
         });
     }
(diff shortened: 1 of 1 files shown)


uxder:
Going to check this in and request a TBR.

src/dom/scroll-render-fix.ts
```diff
@@ -64,11 +64,11 @@ export class ScrollRenderFix {
...
         this.raf.read(()=> {
           this.targetY = document.documentElement.scrollTop + e.deltaY;
-        });
-        this.raf.postWrite(()=> {
-            if (this.currentY !== this.targetY) {
-              document.documentElement.scrollTop = this.targetY;
-            }
+          this.raf.postWrite(()=> {
+              if (this.currentY !== this.targetY) {
+                document.documentElement.scrollTop = this.targetY;
```
angusm, line 69: Looks good. Is there a case where e.target would not be the documentElement?

Does this capture a scroll event on a div element with `overflow: scroll`?

PR #66, opened 2021-01-29
Port the Toolbox Carousel

angusm:
Carousel component.

Has the option to provide separate Transition instances to the carousel to handle different types of interaction.

As part of this pull request two transitions are included, a default transition that only sets before/active/after classes on slides, and a PhysicalSlide carousel that can be dragged around.

Enabling this PhysicalSlide transition is the Draggable component and the CachedMouseTracker. Use of DomWatcher and MouseTracker was attempted as an alternative but the performance suffered noticeably. PhysicalSlide also has a few pieces broken out into different files to try and ease readability.

A few utility functions have been added or updated.
- arrayf.max supports tie breaker functions
- arrayf.min inverts a call to arrayf.max, mainly existing for the readability of the name
- dom.hasVisibleArea and dom.getVisibleArea are used to aria-hide scrolled off slides
- dom.getVisibleDistanceBetweenCenters is used to determine active slides
- TrackedListener is added, eases disposal of listeners by providing an ID when a listener is added, similar to the behaviour of setTimeout.
- sef.subtract and setf.merge are added to ease working with Sets over arrays.


src/arrayf/arrayf.ts
```diff
@@ -32,23 +33,87 @@ export class arrayf {
...
+     * Return value from array that generates the lowest return value when
+     * passed as a parameter to the score function.
+     *
+     * Multiple score functions can be passed with later functions used only to
+     * break ties for values returned from the previous score function.
+     *
+     * Example:
+     * ```
+     * const values = [{'a': 2, 'b': 3}, {'a': 1, 'b': 4}];
+     * arrayf.max(values, (x) => x.a); // Returns {'a': 1, 'b': 4}
```
uxder, line 83: array.max -> array.min

src/dom/dom.ts
```diff
@@ -980,10 +1025,77 @@ export class dom {
         allTextNodes.forEach((node)=> {
            node.nodeValue = node.nodeValue.replace(/\s+([^\s]*)\s*$/, nbsp + '$1')
         })
+    }
 
+    /**
+     * Add an event listener to multiple events.
+     * Useful for when mouse and touch events should call the same listener.
+     */
+    static addEventListeners(
```
uxder, line 1034: I'm okay with adding this here but FYI, I think we should try to use DomWatcher as a general pattern.

src/dom/dom.ts
```diff
@@ -980,10 +1025,77 @@ export class dom {
...
+     * translations that will be applied via MatrixService.
+     *
+     * @param a
+     * @param b
+     */
+    static getVisibleDistanceBetweenCenters(
+        a: HTMLElement, b: HTMLElement = null
+    ): Vector {
+        // Gather up the information on the first element's center position.
+        const aRect = a.getBoundingClientRect();
```
uxder, line 1075: I'm a bit confused why we need MatrixService here given that getBoundingClientRect accounts for translation positions.  Do we need this?

src/dom/dom.ts
```diff
@@ -980,10 +1025,77 @@ export class dom {
...
+            (event) => element.removeEventListener(event, listener));
     }
 
     static getScrollElement(): Element {
         return document.scrollingElement || document.documentElement;
     }
+
+    /**
+     * Determine the distance between the centers of two given elements.
+     * If no second element is given, the root element is used.
```
uxder, line 1062: I think we can just say the root is the document to make it clearer.

src/dom/tracked-listener.ts
```diff
@@ -0,0 +1,74 @@
...
+type Listener = [EventTarget, string, EventListenerOrEventListenerObject];
+
+/**
+ * Used for tracking listeners by ID.
+ * This way event listeners in a component can be tossed into a set and easily
+ * removed during dispose. Otherwise which element had which events tied to
+ * which functions needs to be tracked in the class which can be annoying with
+ * dynamically added listeners.
+ */
+export class TrackedListener {
```
uxder, line 11: Seems like there is a overlap with DomWatcher class.  Is DomWatcher something you can use instead?

src/mathf/mathf.ts
```diff
@@ -858,9 +858,11 @@ export class mathf {
    * @return {number} The interporalated value.
    */
   static lerp(value1: number, value2: number, amount: number): number {
```
uxder, line 860: Yeah.......so this is definitely the correct traditional lerp

(in c#)
```c#
Lerp(float a, float b, float t) {
  return (1.0f - t) * a + b * t;
}
```

but....I can't remember but as commented out, there was some reason why I went with current calculations as I did go back and forth at some point.  

For this reason, I'm a bit worried about regression here given since many classes depend on this working as is.  Do you see any issues?  If not, although not ideal, I would almost ask to leave this as is and make a lerp2.

src/mathf/cubic-bezier.ts
```diff
@@ -75,6 +77,9 @@ import { Vector } from './vector';
  * @see https://stackoverflow.com/questions/27053888/how-to-get-time-value-from-bezier-curve-given-length
  */
 export class CubicBezier {
+    static EASE_IN_OUT_SINE: EasingFunction =
```
uxder, line 80: Do we need this?

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,123 @@
+import { setf } from '../../setf/setf';
+import { Carousel } from './carousel';
+
+/**
+ * Will synchronize the slide positions of multiple carousels.
+ *
+ * This can be useful if you have separate areas for assets, copy and/or
+ * navigation that all need to be synchronized together.
+ */
+export class CarouselSynchronizer {
```
uxder, line 10: I'm a bit confused and not following how this might used in practice?  Are you provide an example?

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,123 @@
+import { setf } from '../../setf/setf';
+import { Carousel } from './carousel';
+
+/**
+ * Will synchronize the slide positions of multiple carousels.
+ *
+ * This can be useful if you have separate areas for assets, copy and/or
+ * navigation that all need to be synchronized together.
+ */
```
uxder, line 9: Maybe you can add the example you have in examples/carousel.html to make this clearer.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,508 @@
...
+  readonly element: HTMLElement;
+  readonly drivenBySync: boolean;
+
+  constructor(element: HTMLElement, drivenBySync: boolean = false) {
+    this.element = element;
+    this.drivenBySync = drivenBySync;
+  }
+}
+
+export class Carousel {
```
uxder, line 56: More of an overall comment on the Carousel structure, I usually prefer a simple structure for carousels and UI components (as you may have seen on some of our projects) - usually a single class without much cross dependency.

There seems to be a lot of manual calculations (linear alg, vector, matrices) which I think when we are doing things fairly complicated like a super custom hero or webGL stuff but where possible for something like a carousel, I think it might be best to just keep it simple and learn on the browser to do it's work.

However, this might be because your carousel has provides advantages that I don't quite see yet.  Can you help me understand how `adjust-slide-for-loop`, `adjust-slide-for-split`, `physical-slide` all kind of fit together here?

uxder (changes requested):
Thanks Angus!  I left a few notes.

src/mathf/mathf.ts
angusm: It looks like this version wasn't used previously, the initial commit is:
```

  /**
   * Linear interpolate from start to end given amount.
   * ```ts
   *
   * ```
   * @param {number} value1 The low-end of the range to lerp.
   * @param {number} value2 The high-end of the range to lerp.
   * @param {number} amount A value between 0-1.  1 would essentially mean no lerp.
   */
  static lerp(value1: number, value2: number, amount: number) {
    amount = amount < 0 ? 0 : amount;
    amount = amount > 1 ? 1 : amount;
    return value1 + (value2 - value1) * amount;
  }
```

The only concern would be if we have something that depends on how it currently handles negatives, but I'm not sure if that's something we should support?

I'll add a couple more tests that cover the issue this addresses.

src/mathf/cubic-bezier.ts
angusm: No, I thought it might be handy to start adding cheat-sheet style constants, but I can move this inline to where it's used.

src/mathf/mathf.ts
uxder: Yeah I think if it passes the tests in mathf.spec.ts it might be okay then.  :)

src/ui/carousel/carousel.ts
angusm: Of course. I prefer simpler too and I'll see what else I can streamline or incorporate tighter with existing Degu code.

That said, this started much simpler and more akin to what we've seen in other projects. First lines were committed back in March of 2018, but the complexity has been built up to address corner cases and performance issues encountered across projects.

This carousel for instance, needed transitions based on slide position updated in real time. That's where MatrixService played in, updating the scale of the device with transform information that would be changed by the time the scale change was rendered resulted in a jitter. This necessitated referencing translation changes made in the read step, to perform other calculations before making adjustments during the write step. https://store.google.com/us/product/google_nest_mini#mini-solutions-audio

`adjustSlideForLoop` fixes the need for duplicating slides in an infinite loop situation, allowing for slides to transform independently outside of their container. I know the example didn't reflect this, I found the bug during the port and its been rectified. In short, if a slide has moved far enough offscreen to either side it puts it on the opposite side. I'll see if I can find a better name and better document it in code.

`adjustSlideForSplit` Ensures that other slides keep the appropriate distance from the active slide.

`physicalSlide` handles user interaction for dragging slides. The name should probably be updated now that I'm thinking about it. At one point, probably years ago now, it had friction/velocity/etc. and was "physical" that way, which it isn't anymore.

I'll take another polish pass through and see if I can better explain/justify what exists and see if there's anything further I can remove without hurting performance.

src/ui/carousel/carousel-synchronizer.ts
angusm: Done.

src/dom/tracked-listener.ts
angusm: Done.

src/dom/dom.ts
angusm: Put `document.children[0]`to be explicit to avoid any confusion.

src/dom/dom.ts
angusm: I'll move this to within the carousel. Sorry, at one point MatrixService lived outside of Carousel, when it was consolidated this was missed.

The reason MatrixService is used is so that when this function is queried it takes into account translations that are scheduled to be applied during the next write step. Without this some effects become out of sync and jitter.

src/dom/dom.ts
angusm: Removed. Used DomWatcher.

src/arrayf/arrayf.ts
angusm: Done.

angusm:
Fixed where straightforward, responded where less straightforward.

Taking another pass to see what else I can remove/simplify/inline.

angusm:
Hi Scott,

I've attempted to remove/inline/rename code in the hopes of making it a bit easier to follow.

Would you mind taking another look when you get a chance?

src/ui/draggable/draggable-synchronizer.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export class DraggableSynchronizer {
+  /**
+   * Returns a singleton. `use` is passed to ensure the singleton is only
+   * actually disposed when it is done being used everywhere. This same `use`
+   * should be passed to the dispose call.
+   * @param use
+   */
+  static getSingleton(use: any): DraggableSynchronizer {
+    DraggableSynchronizer.singletonUses.add(use);
```
uxder, line 14: Can you tell me what the case for singletonUses is?  I'm wondering if we need it.

src/mathf/mathf.spec.ts
```diff
@@ -84,6 +84,10 @@ test('lerp', t => {
     t.is(mathf.lerp(0, 100, 0.5), 50);
     t.is(mathf.lerp(25, 79, 0.2), 35.8);
 
+    t.is(mathf.lerp(-25, 50, 0.5), 12.5);
```
uxder, line 87: Thanks for adding :)

src/ui/draggable/draggable-synchronizer.ts
angusm: It's a performance/disposal measure.

A singleton for synchronizing ensures that if we have two sync'd sets of Draggables: [a, b, c] and [d, e, f], if e and b are later synchronized, then the entire set [a, b, c, d, e, f] are synchronized.

Whereas without singletons we'd have [a, b, c], [d, e, f] and [e, b] which could result in jitter.

Where the singletonUses comes in is to ensure that the singleton is not discarded as soon as one use of the Singleton is disposed.

Without tracking uses, if we have Draggables T and U, they would both call upon getSingleton() to get a DraggableSynchronizer instance. If T then disposed itself and its DraggableSynchronizer instance, U would then be using a disposed instance of DraggableSynchronizer.

Whereas with uses being tracked, the singleton isn't fully disposed until all instances that use it have indicated they are done.

This pattern repeats itself in CachedMouseTracker and the CarouselSynchronizer.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,506 @@
...
+    const slidesAfter = this.getSlidesAfter(activeSlide);
+
+    const adjustCssClasses =
+        (slide: HTMLElement, cssClassesToKeep: Set<string>) => {
+          const currentCssClasses = this.slideCssClasses.get(slide);
+          this.slideCssClasses.set(slide, cssClassesToKeep);
+          const classesToRemove =
+              setf.subtract(currentCssClasses, cssClassesToKeep);
+          slide.classList.remove(...classesToRemove);
+          slide.classList.add(...cssClassesToKeep);
```
stevenle, line 460: Maybe I'm misunderstanding what these lines do, but instead of allocating sets and calculating differences, couldn't you do something like:

```
el.className = classesToKeep; // as string
```

src/ui/carousel/draggable-slide/matrix-service.ts
```diff
@@ -0,0 +1,103 @@
+import { Matrix } from './matrix';
+import { DefaultMap } from '../../../map/default-map';
+import { Raf } from '../../..';
+import { Vector } from '../../../mathf/vector';
+
+/**
+ * This service is used to coordinate modifications to a transform coming
+ * from multiple services. This allows us to stack transform-effecting UI
```
stevenle, line 8: Do we have any existing examples where two different pieces of code are modifying a single element's transforms?

src/ui/carousel/draggable-slide/slide-to-draggable-map.ts
```diff
@@ -0,0 +1,57 @@
...
+  const totalSlideWidth =
+      mathf.sum(slides.map((s) => s.offsetWidth));
+  const lastSlideWidth = slides.slice(-1)[0].offsetWidth;
+  const halfLastSlide = lastSlideWidth / 2;
+  const halfFirstSlide = slides[0].offsetWidth / 2;
+
+  const min = halfContainer - totalSlideWidth + halfLastSlide;
+  const max = halfContainer - halfFirstSlide;
+  const currentX =
+      Matrix.fromElementTransform(draggable.element).getTranslateX();
```
stevenle, line 32: Is there an easier way to get an element's X transform without having to create a Matrix()?

src/ui/carousel/draggable-slide/slide-to-draggable-map.ts
```diff
@@ -0,0 +1,57 @@
...
+  const deltaX = finalX - currentX;
+
+  return new Vector(deltaX, delta.y);
+}
+
+/**
+ * Map slide elements to Draggables that are appropriately constrained to the
+ * carousel.
+ */
+export class SlideToDraggableMap extends DefaultMap<HTMLElement, Draggable> {
```
stevenle, line 43: Where is this class being used?

This also feels highly coupled to the carousel implementation, I would maybe suggest moving this into carousel.ts as a non-exported member.

Alternatively, can this be implemented as a map within Carousel itself?

src/ui/draggable/draggable-synchronizer.ts
uxder: Hmmm...does this mean we have multiple singletons?  By definition of a singleton, I don't think this is a good pattern and confused by we have multiple singletons.

Can we just not do:

new CarouselSynchronizer();

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,506 @@
...
+    this.allowLooping = allowLooping;
+    this.condition = condition;
+    this.container = container;
+    this.lastActiveSlide = null;
+    this.onTransitionCallbacks = onTransitionCallbacks;
+    this.onDisposeCallbacks = onDisposeCallbacks;
+    this.slides = slides;
+    this.transition = transition !== null ? transition : new CssClassesOnly();
+    this.transitionTarget = null;
+    this.synchronizer = CarouselSynchronizer.getSingleton(this);
```
uxder, line 123: I think overall the pattern of components looking for instances of singletons can be confusing to follow and makes architecture a harder to follow.

In general, I find it easier if components are self composed and follow a top down composition.  

Here you have a case of Carousel, doing horizontal lookup of CarouselSynchronizer and then later calling methods of CarouselSynchronizer.  I'm not exactly sure what sync is doing here but it feels a bit coupled.

Can we architect this in a different way?

I think instead this might be easier to follow if the carousel was solely responsible to updating itself.  It would have a render() method.  It should have no knowledge of what a CarouselSynchronize is.  You could have a public method that perhaps disabled the internal Raf like stopLoop().

You then have multiple instances of Carousels.

You instantiate a CarouselSynchronizer([Carousels...]).  CarouselSynchronizer would call stopLoop() on each carousel component and then run it's own raf and run the render loop on each Carousel.

Perhaps this is not exactly what you are doing but the point being, I'm wondering if we can decouple this more.

src/ui/draggable/draggable.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  /**
+   * Disposes of the Draggable instance.
+   */
+  dispose() {
+    this.raf.stop();
+    this.constraints = [];
+    this.mouseTracker.dispose(this);
+    this.draggableSynchronizer.dispose(this);
```
uxder, line 53: See my comment on Carousel and how I think we should avoid this pattern of multiple singletons.

src/ui/draggable/horizontally-draggable.ts
```diff
@@ -0,0 +1,21 @@
+import { Draggable, DraggableConstraint } from './draggable';
```
uxder, line 1: I think in terms of library consistency, looking at the way I wrote the other components rather than extending Draggable, we would just make an config in Draggable like this.

I personally think this is much easier to follow for other developers.

```
new Draggable({
   "horizontal": true
})
```

I think for the most part, for these simpler components, I prefer keeping it simple than being too quick to subclass which can lead to prematurely complicating it.

uxder (commented):
Hey Angus,

I had a few more comments on some higher level architecture for this work.

Thanks for all your work:)

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,506 @@
...
+  BEFORE_SLIDE = 'before',
+  AFTER_SLIDE = 'after'
+}
+
+/**
+ * Tracks an element that should be transitioned to.
+ */
+class TransitionTarget {
+  readonly element: HTMLElement;
+  readonly drivenBySync: boolean;
```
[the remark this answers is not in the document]
angusm: Added more comments. Just a small data structure for grouping related variables.

src/ui/carousel/draggable-slide/matrix.ts
```diff
@@ -0,0 +1,78 @@
...
+    return new Vector(this.tx, this.ty);
+  }
+
+  translate(vector: {x: number, y: number}): Matrix {
+    const newX = this.tx + vector.x;
+    const newY = this.ty + vector.y;
+    return new Matrix(this.a, this.b, this.c, this.d, newX, newY);
+  }
+
+  setPosition(position: {x: number, y: number}): Matrix {
```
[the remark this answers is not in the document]
angusm: Yes I can make this adjust, it was typed this way to allow easy passing of degu Vector objects without making them necessary. But (x, y) is just as good and more readily understandable so I'll make the change.

I really wish typescript had overloading.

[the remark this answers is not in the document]
stevenle: fwiw it does have overloading, you can provide different method signatures and one implementation, e.g.:

```typescript
function add(a: string, b: string): string;
function add(a: number, b: number): number;
function add(a: any, b: any): any {
    return a + b;
}
```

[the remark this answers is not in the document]
stevenle: in your case, i think you can maybe do something like:

```typescript
setPosition(vector: Vector): Matrix;
setPosition(x: number, y: number): Matrix;
setPosition(x: number | Vector, y?: number): Matrix {
  ...
}
```

that said, i do think a regular `setPosition(x, y)` is an easier api to maintain.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,506 @@
...
+    this.transitionTarget = null;
+    this.synchronizer = CarouselSynchronizer.getSingleton(this);
+
+    this.slideCssClasses =
+        DefaultMap.usingFunction<HTMLElement, Set<string>>(
+            () => new Set<string>());
+
+    /** Mapped Set caches to avoid allocation churn */
+    this.activeCssClassSet = new Set([activeCssClass]);
+    this.beforeCssClassMap =
```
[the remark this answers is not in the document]
angusm: I'm sorry, I don't follow. This caching is designed to set the value once.

These are to prevent allocating memory for the same objects multiple times, so they reduce overall memory.

I need the sets so items can be quickly checked for membership, I could do it with arrays if they are less memory intensive than sets, but we're giving up O(1) access for O(n) access.

What I can do is drop the activeCssClassSet but:
- It would prevent the same method from being used for changing the active slides classes that is used to change the other slides classes due to the different data structure.
- A set would still need to be created for it when setting this.slideCssClasses.

src/ui/carousel/carousel.ts
angusm: This would prevent other code from being able to work with the classlist of the same element.

angusm: The extra tracking of class names in sets in this class is so that the carousel only removes and adds its own classes. If we weren't concerned with other potential code modifying the element we could use the `el.classList` features instead but we risk creating side effects.

src/ui/carousel/draggable-slide/matrix-service.ts
angusm: Yes. But not that I can post in the public repo.

src/ui/carousel/draggable-slide/slide-to-draggable-map.ts
angusm: Yes. We could parse only the X value from the result of `dom.getComputedStyle(el).transform`

The effort here was to create a canonical way of grabbing transform information so there would be a single point to update if browser quirks were introduced. However if we assume native functionality to be reliable I may be able to trim this Matrix content. I will investigate.

src/ui/carousel/draggable-slide/slide-to-draggable-map.ts
angusm: This is used in DraggableSlide and works best as a class, but our current linter is demanding one class per file. Should we remove that lint rule?

src/ui/carousel/carousel.ts
angusm: I'll take a look. This has always bothered me as well, one of those things that survived because it was fast to implement and then didn't pose enough problems to warrant priority later.

src/ui/draggable/draggable-synchronizer.ts
angusm: No, it is still a single singleton. We just can't rely on any user of the singleton to know fully when all other users of the singleton are done with it in order to dispose of it. So we can either have the singleton exist in perpetuity once instantiated, or we have it track its users so it is able to self-dispose once it's last user has indicated it is no longer needed.

If it is not a singleton it cannot be aware of all instances in which a carousel was synchronized and we could have multiple synchronizers playing tug of war with a single carousel.

src/ui/draggable/draggable.ts
angusm: Responded in the carousel comment.

examples/carousel.html
```diff
@@ -0,0 +1,89 @@
+
+
```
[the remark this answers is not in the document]
angusm: Done.

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,208 @@
...
+ * </div>
+ * <div class="nav">
+ *   <div class="dot">1</div>
+ *   <div class="dot">2</div>
+ *   <div class="dot">3</div>
+ * </div>
+ * ```
+ *
+ * JS:
+ * ```
```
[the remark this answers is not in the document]
angusm: Done.

examples/carousel.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * This sample show the most basic uses of CanvasImageSequence which is
```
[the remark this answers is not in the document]
angusm: Done.

examples/carousel.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+import {Carousel} from '../lib/ui/carousel/carousel';
+import {DraggableSlide} from '../lib/ui/carousel/draggable-slide/draggable-slide';
+import {CarouselSynchronizer} from '../lib/ui/carousel/carousel-synchronizer';
+
+export default class CarouselSample {
+  constructor() {
+    const a = new Carousel(
+        document.querySelector('.carousel'),
+        Array.from(document.querySelectorAll('.carousel .slide')),
```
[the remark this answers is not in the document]
angusm: It does not, but it could be added.

src/dom/dom.ts
```diff
@@ -880,7 +879,51 @@ export class dom {
...
 
+    /**
+     * Determine if the target element has a visible area within the given
+     * container element.
+     * @param target
+     * @param container
+     */
+    static hasVisibleArea(
+        target: HTMLElement,
+        container: HTMLElement = null
```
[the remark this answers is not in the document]
angusm: Good catch, no it should not be.

In Toolbox this method would default to just checking within the root element, but I tried to simplify the port and missed this.

src/dom/dom.ts
```diff
@@ -880,7 +879,51 @@ export class dom {
         return isDisplayNone;
     }
 
+    /**
+     * Determine if the target element has a visible area within the given
+     * container element.
+     * @param target
+     * @param container
+     */
+    static hasVisibleArea(
```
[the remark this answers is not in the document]
angusm: Looks like other streamlining has made this unnecessary, yanked the method altogether.

src/setf/setf.ts
```diff
@@ -0,0 +1,28 @@
+export class setf {
+  /**
+   * Return a new set containing all the values in the minuend that do not
+   * appear in the subtrahend.
+   */
+  static subtract<T>(minuend: Set<T>, subtrahend: Set<T>): Set<T> {
```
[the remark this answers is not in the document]
angusm: Good call done.

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,208 @@
...
+        syncedCarousel.transitionToIndex(index, true);
+      }
+    });
+  }
+
+  /**
+   * Dispose of the given carousel, removing it from any synced sets.
+   * @param carousel
+   */
+  disposeCarousel(carousel: Carousel) {
```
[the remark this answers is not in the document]
angusm: Renamed.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,506 @@
...
+enum Direction {
+  LEFT = -1,
+  RIGHT = 1
+}
+
+/**
+ * Options provided to the constructor
+ */
+export interface CarouselOptions {
+  onTransitionCallbacks?: Array<(carousel: Carousel) => void>;
```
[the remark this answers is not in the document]
angusm: Done.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,506 @@
...
+    this.beforeCssClass = beforeCssClass;
+    this.afterCssClass = afterCssClass;
+    this.allowLooping = allowLooping;
+    this.condition = condition;
+    this.container = container;
+    this.lastActiveSlide = null;
+    this.onTransitionCallbacks = onTransitionCallbacks;
+    this.onDisposeCallbacks = onDisposeCallbacks;
+    this.slides = slides;
+    this.transition = transition !== null ? transition : new CssClassesOnly();
```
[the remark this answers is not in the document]
angusm: Done.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,506 @@
...
+          this.transition.hasTransitionedTo(this.transitionTarget.element);
+
+      const shouldSync = !this.transitionTarget.drivenBySync;
+      if (hasTransitionedToTarget) {
+        this.transitionTarget = null;
+      } else {
+        this.transition.transition(this.transitionTarget.element);
+      }
+      return shouldSync;
+    } else {
```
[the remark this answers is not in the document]
angusm: Done.

src/ui/draggable/draggable.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.domWatcher.dispose();
+  }
+
+  /**
+   * Start a drag.
+   */
+  protected startInteraction(): void {
+    this.raf.postWrite(() => { // Trigger after position has been updated
+      if (this.isInteracting()) { return; }
```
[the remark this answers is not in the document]
angusm: Fixed.

src/ui/carousel/draggable-slide/matrix.ts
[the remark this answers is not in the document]
angusm: Done.

src/ui/carousel/draggable-slide/matrix.ts
```diff
@@ -0,0 +1,78 @@
...
+
+const STYLE_STRING_PREFIX = 'matrix(';
+const STYLE_STRING_PREFIX_LENGTH = STYLE_STRING_PREFIX.length;
+
+/**
+ * Tracks information on an element's transform.
+ */
+export class Matrix {
+  static parseFromString(str: string): Matrix {
+    if (str === 'none' || !str.length) {
```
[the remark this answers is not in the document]
angusm: Ran the benchmarks. Switched order but kept `.length`.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,506 @@
...
+  private readonly beforeCssClass: string;
+  private readonly beforeCssClassMap: DefaultMap<number, Set<string>>;
+  private readonly afterCssClass: string;
+  private readonly afterCssClassMap: DefaultMap<number, Set<string>>;
+  private readonly condition: () => boolean;
+  private readonly container: HTMLElement;
+  private readonly slides: HTMLElement[];
+  private readonly transition: Transition;
+  private readonly allowLooping: boolean;
+  private readonly onTransitionCallbacks: Array<(carousel: Carousel) => void>;
```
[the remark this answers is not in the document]
angusm: The linter prefers Type[] for simple types, but if it's more complex Array<(f: Foo) => Bar> is preferred.

src/ui/draggable/horizontally-draggable.ts
angusm: Done.

src/ui/draggable/draggable.ts
```diff
@@ -0,0 +1,150 @@
...
+ */
+export enum DraggableEvent {
+  START = 'deguDraggableStart',
+  END = 'deguDraggableEnd'
+}
+
+/**
+ * Makes a DOM element draggable by adjusting the element's transform.
+ */
+export class Draggable {
```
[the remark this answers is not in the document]
angusm: Yes, this should support non-carousel use.

I will get a demo setup.

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,208 @@
...
+    CarouselSynchronizer.singletonUses.add(use);
+    if (CarouselSynchronizer.singleton === null) {
+      CarouselSynchronizer.singleton = new CarouselSynchronizer();
+    }
+    return CarouselSynchronizer.singleton;
+  }
+
+  private static singleton: CarouselSynchronizer = null;
+  private static singletonUses: Set<any> = new Set();
+  private syncedCarousels: Array<Set<Carousel>>;
```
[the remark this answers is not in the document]
angusm: I can give it a shot. Going to attempt to re-work how tightly coupled this is per Scott's comment so this can be part of that rewrite.

src/ui/carousel/carousel.ts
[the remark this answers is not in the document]
stevenle: I took some time to trace what's going on, and my understanding is that all these sets are being allocated so we know which elements have classes of `active` vs `before before--{index}` and `after after--{index}`.

Sets are extremely memory intensive and I think you can simplify this logic so that it's more readable. You can do something like:

on active element: remove "before" and "after" classes on the element. apply "active"
on inactive elements: remove "active" class. apply either "before" or "after" class.
on all elements: set something like `data-index="{index}"`

This would avoid us having to store and figure out what classes we've already set on the slide.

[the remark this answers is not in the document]
angusm: Done. I hadn't considered using an attribute for specifying the index, good call.

src/ui/carousel/carousel.ts
angusm: Resolved with attribute setup suggested in other comment.

src/ui/carousel/draggable-slide/slide-to-draggable-map.ts
angusm: Inlined the class.

src/ui/carousel/carousel-synchronizer.ts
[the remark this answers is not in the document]
angusm: Done.

src/ui/draggable/draggable-synchronizer.ts
angusm: Re-worked this using the new graph setup Steven suggested. Disposal is no longer called, uses are no longer tracked, instead `removeCarousel` must be called where CarouselSynchronizer is used and it self-disposes as below:

```

  /**
   * Remove the given carousel from any synced sets.
   * @param carousel
   */
  removeCarousel(carousel: Carousel) {
    const syncedCarousels = this.carouselGraph.get(carousel);
    syncedCarousels.forEach((syncedCarousel) => {
      this.carouselGraph.get(syncedCarousel).delete(carousel);
    });
    this.carouselGraph.delete(carousel);

    if (
        this.carouselGraph.size === 0 &&
        CarouselSynchronizer.singleton === this
    ) {
      CarouselSynchronizer.singleton = null;
    }
  }
  ```

File: src/ui/carousel/carousel.ts (hunk 17 above)
Comment (angusm) [reply src/ui/carousel/carousel.ts]: Removed reference to CarouselSynchronizer from within carousel.ts.

File: src/ui/carousel/draggable-slide/slide-to-draggable-map.ts (hunk 15 above)
Comment (angusm) [reply src/ui/carousel/draggable-slide/slide-to-draggable-map.ts]: Updated to use the MatrixService so it is grabbing the cached value.

So I re-read my code to remind myself why I had this complexity.

The MatrixService allows the dragging code to be separate from the slide/carousel specific code.

Without the MatrixService the code that loops slides asteroid-style overwrites the transformation applied by the Draggable code. It would also overwrite any styles on the element, though that may be less of a concern and any issue this creates could probably be resolved by just nesting another div.

So I could consolidate this all into a single DraggableSlide piece and remove the standalone Draggable and the need to have the Matrix/MatrixService if that's preferable, then deal only with the X component of the matrix and set a single inline style for `translateX()` rather than an inline style for the Matrix that preserves initial transforms set via CSS or other code acting on the element.

**Comment by angusm** [comment]
Made additional changes in response to the comments.

Current area of concern is finding out what folks thoughts are on the tradeoffs of keeping Draggable and Slide behaviour separate versus consolidating, my response to Steven has a bit more context.

Also left the feedback on CarouselSynchronizer disposal open to make sure the change I made is acceptable.

File: src/ui/carousel/draggable-slide/slide-to-draggable-map.ts (hunk 15 above)
Comment (uxder) [reply src/ui/carousel/draggable-slide/slide-to-draggable-map.ts]: Maybe dumb question but is this solved by adding say adding an inner div and adding one transform to the root and the other child?  If that is the case, I think I prefer adding the extra element over trying to solve matrix calculation conflicts to simply this...

(that being said, I think the matrix service class might be useful in cases where this is unavoidable so I think it would be nice to have that in degu eitherway)

File: examples/carousel.js [hunk 39]
```diff
@@ -24,5 +24,12 @@ export default class CarouselSample {
           allowLooping: true,
         });
     CarouselSynchronizer.getSingleton(this).sync(a, b);
+    const c = new Carousel(
+        document.querySelector('.carousel--c'),
+        Array.from(document.querySelectorAll('.carousel--c .slide')),
+        {
+          allowLooping: false,
+          transition: new DraggableSlide(),
```
uxder, line 32: No action needed.

For future, It'd be awesome if you can expand the examples with other transitions and other settings.

src/arrayf/arrayf.ts
```diff
@@ -32,23 +33,96 @@ export class arrayf {
...
      * passed as a parameter to the score function.
      *
+     * Multiple score functions can be passed with later functions used only to
+     * break ties for values returned from the previous score function.
+     *
      * Example:
      * ```
      * const values = [{'a': 2, 'b': 3}, {'a': 1, 'b': 4}];
      * arrayf.max(values, (x) => x.a); // Returns {'a': 2, 'b': 3}
      * arrayf.max(values, (x) => x.b); // Returns {'a': 1, 'b': 4}
```
uxder, line 43: Could you add an arrayf.spec file and add these as tests?

degu doesn't have tests for class that directly manipulate dom but for low level methods, I try to add a few tests.

src/ui/carousel/draggable-slide/slide-to-draggable-map.ts
angusm: Yes, nesting the inner div solves this conflict.

The only advantage to nesting these divs would be it letting us remove the need for MatrixService, but if we want to keep MatrixService there's not as much benefit to streamlining to only `translateX()` styles.

angusm:
Reworked with the following assumptions:
- Draggable code does not need to be re-used in DraggableSlide
- MatrixService is not necessary and Carousel's DraggableSlide transition is safe to override other transforms by in-lining a transform style.

This has resulted in a good chunk of deleted code :)

I moved away from queryselectors for passing the container/slides to the carousel so that partials built in grow won't need a unique identifier to scope the selector if the partial is duplicated.

stevenle (commented):
Been thinking more about how to simplify the API here. Left two suggestions, I'll keep thinking about ways we can improve the API for end user ease of use.

I know this review is getting quite long, one suggestion might be to move this into experimental temporarily until we lock down the API and can demonstrate its use on a few projects.

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,187 @@
...
+ * .dot.before, .dot.after {
+ *   transform: scale(0.9);
+ * }
+ * ```
+ */
+export class CarouselSynchronizer {
+  /**
+   * Returns a singleton.
+   */
+  static getSingleton(): CarouselSynchronizer {
```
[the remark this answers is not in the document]
angusm: The change is doable, but the Carousel class is already quite large.

The more behaviour we encapsulate in a single class the harder it may become to debug the class in isolation. As it stands, if there is an issue with a carousel  syncing, we can remove the sync and test the carousel in isolation, narrowing down which file we need to be concerned with.

If the singleton pattern is the primary issue we can ditch it and do `new CarouselSynchronizer(...carousels)` instead, without giving up the debug/maintenance advantages of the separation of concerns.

2 or 3 would probably be the common case, Assets/Dots/Copy being synced. Some modules may need more, I have a current example that is synchronizing 7 carousels.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,527 @@
...
+  transitionToIndex(index: number, drivenBySync: boolean = false): void {
+    const clampedIndex = this.getClampedIndex(index);
+    this.transitionToSlide(this.getSlideByIndex(clampedIndex), drivenBySync);
+  }
+
+  /**
+   * Register a function to be called when the carousel starts a transition.
+   * @param callback
+   */
+  onTransitionStart(callback: (carousel: Carousel) => void) {
```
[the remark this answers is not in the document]
uxder: No establish pattern since there aren't many UI components.

lazy-image and a few others do fire events on the root dom element itself.  Example: https://github.com/blinkkcode/degu/blob/main/src/dom/lazy-video.ts#L205

src/ui/carousel/carousel-synchronizer.ts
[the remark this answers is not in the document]
uxder: Per side chat with Angus, I think the case to have to sync multiple levels of carousels is quite limited so I personally prefer remove the singleton usage and instead go with multiple instances of CarouselSynchronizer.

I think if the Carousel class is substantially large, we can just keep as a separate class.

uxder (commented):
Hey Angus,

Thanks for all the hard work.  I think this most recent changes have made it much clearer that I think looks good.   The docs and clear config also really help to just pick this quickly so I really appreciate that.

src/ui/carousel/carousel-synchronizer.ts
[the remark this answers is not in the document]
angusm: Kept CarouselSynchronizer separate, removed singleton requirement.

src/ui/carousel/carousel.ts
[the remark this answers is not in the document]
angusm: Setup EventManager class and EventDispatcher interface per our chat.

angusm:
Hi Scott/Steven,

I've updated the CarouselSynchronizer and events on Carousel :) Please let me know your thoughts when you have a moment.

examples/carousel.js
```diff
@@ -0,0 +1,53 @@
...
+            easingFunction: easeOutBounce
+          }),
+        });
+    const b = new Carousel(
+        document.querySelector('.nav'),
+        Array.from(document.querySelectorAll('.nav .dot')),
+        {
+          allowLooping: true,
+        });
+    new CarouselSynchronizer(a, b);
```
stevenle, line 44: Hmm, so I don't love having to create a new object like this to synchronize two carousels. Can we create a helper function within carousel that handles this object creation? Something like:

```
carouselA.sync(carouselB);
```

Also it'd be helpful for the carousel to store some reference to the `CarouselSyncronizer` so that it can remove itself when it gets `dispose()`d.

angusm: Added a shorthand `sync(...carousels)` method.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,536 @@
...
+   * Return the current transition target.
+   */
+  getTransitionTarget(): TransitionTarget {
+    return this.transitionTarget;
+  }
+
+  /**
+   * Synchronize this carousel with the given carousels.
+   */
+  sync(...carousels: Carousel[]): void {
```
stevenle, line 394: It seems like `this.carouselSynchronizer` is only set on the carousel that `sync()` is called on. e.g.:

```
a = new Carousel(...)
b = new Carousel(...)
a.sync(b)
a.carouselSynchronizer  // -> Object
b.carouselSynchronizer  // -> null/undefined
```

We also need some deterministic way of ensuring that any combination of `sync()` calls trigger updates to every linked carousel. I dunno if this class is easy to unit test but I'd want to ensure something like this works:

```
a.sync(b)
c.sync(d)
b.sync(c)
```

We can potentially simplify the logic of creating sync objects if we assume only 2 carousels are synced at a time (e.g. changing the method signature to `sync(carousel: Carousel)`). Then we can do something like:

* when syncing `a` and `b`, check if a sync object exists on `a` or `b` and use that if so
* if both have a `sync` object, either find some way of merging the two objects so they're linked  or take a lazy approach and just throw an error
* if no sync object exists on either, create a new sync object and add that to both `a` and `b`

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,536 @@
...
+  autoplaySpeed: number;
+  private readonly activeCssClass: string;
+  private readonly beforeCssClass: string;
+  private readonly afterCssClass: string;
+  private readonly distanceToActiveSlideAttr: string;
+  private readonly condition: () => boolean;
+  private readonly container: HTMLElement;
+  private readonly slides: HTMLElement[];
+  private readonly transition: Transition;
+  private readonly allowLooping: boolean;
```
stevenle, line 85: Can we call this `loop` instead?

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,173 @@
...
+   *
+   * @param carousel
+   */
+  handleCarouselTransition(carousel: Carousel) {
+    const transitionTarget = carousel.getTransitionTarget();
+    if (!transitionTarget || transitionTarget.drivenBySync) {
+      return; // Skip if this sync was triggered by a sync call
+    }
+
+    const syncedCarousels = this.graph.get(carousel);
```
stevenle, line 146: I think you may need to walk all the linked nodes in the graph since I think the following example would fail:

```
a.sync(b)
c.sync(d)
b.sync(c)
```

^ the above would produce a graph like:
```
graph = {
  a: [b]
  b: [a, c]
  c: [b, d]
}
```

If the following is called:

```
c.transitionToIndex(2)
```

Then only `b` and `d` would be updated but not `a`.

You can sorta conceptualize this example using islands connected by bridges. Imagine you start on an island and need to send a message to every connected island once and only once by crossing bridges. You'll need to keep track of which islands have already been visited and avoid crossing bridges that lead to those islands.

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,173 @@
...
+ *   transform: scale(0.9);
+ * }
+ * ```
+ */
+export class CarouselSynchronizer {
+  private graph: DefaultMap<Carousel, Set<Carousel>>;
+
+  constructor(...carousels: Carousel[]) {
+    this.graph = DefaultMap.usingFunction((carousel: Carousel) => new Set());
+    this.sync(...carousels);
```
stevenle, line 108: Maybe it's just the way I think about how object creation should work, but in my mind the constructor should basically set up the attributes for the class and shouldn't necessarily do the actual work. So I'd suggest the example usage of this class be something like:

```
const syncer = new CarouselSynchronizer(); 
syncer.sync(a, b); 
```

I think this helps readability because it tells me "line 1: create a syncer object, line 2: sync carousels a and b". (Whereas when I read `new CarouselSynchronizer(a, b)` I get confused about what's going on.)

(tl;dr: I'm suggesting removing the `carousels` args in the constructor.)

stevenle (commented):
Thanks for all the changes so far, and sorry I keep adding these comments in late since I know I could've/should've pointed out some of these things in earlier comments. 

With the goal of providing a dead simple API to users using this class, I left a few new comments. LMK if anything needs clarification, and of course you're welcome to disagree with anything of the suggestions if you feel the current way works well.

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,173 @@
...
+   * Handle the transition of the given carousel to the given index.
+   *
+   * Will transition all other carousels synced with the given carousel to the
+   * given index.
+   *
+   * @param carousel
+   */
+  handleCarouselTransition(carousel: Carousel) {
+    const transitionTarget = carousel.getTransitionTarget();
+    if (!transitionTarget || transitionTarget.drivenBySync) {
```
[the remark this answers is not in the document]
angusm: null is the lack of a target, it indicates there is no current transition occurring

src/ui/carousel/carousel.ts
angusm: If I'm understanding the concerns correctly, these situations are already handled.

Yes only one CarouselSynchronizer instance is created for a call to sync, this class already handles bi-directional changes from the other carousels.

We can't assume only 2 carousels are synced, I currently have a use case involving 7 carousels.

I can definitely add some demos. I can probably add some unit tests, I'm not sure if we have stuff setup for testing with DOM, I can dig in.

The current big limitation is that if you have `a, b, c` and `sync(a, b)` and `sync(b, c)` a change in a will not prompt a change in c and vice-versa. This was solved by the singleton and it was agreed that allowing this case was acceptable in exchange for removing the singleton pattern.

src/ui/carousel/carousel.ts
angusm: No, it would conflict with the internal loop() method that handles RAF checks. But it shouldn't read as a verb, I will change to allowsLooping so it sounds more like a property.

src/ui/carousel/carousel.ts
stevenle: Right so what I'm suggesting is create a single sync object that's shared between the two carousels so that `a.carouselSynchronzer` and `b.carouselSynchronizer` both exist. And if you do `b.sync(c)` then take the synchronizer from `b` and set it to `c`.

src/ui/carousel/carousel-synchronizer.ts
angusm: I'm good with that.

src/ui/carousel/carousel-synchronizer.ts
angusm: Looks like I made a mis-step somewhere in the conversion to graphs rather than sets. I'll revisit the implementation, this case was definitely handled previously so it should be again.

src/ui/carousel/carousel.ts
angusm: If we have `a, b, c, d`

`a.sync(b)` All good

`c.sync(d)` All good

`b.sync(c)` How should we handle this case? Should we handle this case?

The more we've changed this synchronizer the more cases I feel we're ignoring and it's starting to feel like a step backwards.

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,173 @@
...
+      if (!this.graph.has(carousel)) {
+        carousel.on(
+            CarouselEvent.TRANSITION_START,
+            () => this.handleCarouselTransition(carousel));
+      }
+
+      const set = this.graph.get(carousel);
+      carousels
+          .filter((c) => c !== carousel)
+          .forEach((c) => set.add(c));
```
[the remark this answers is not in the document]
angusm: Fair enough. I'll aim to prefer this layout going forward.

Previously worked with folks who pushed chaining functional-style one-liners but I'm not fussed :)

src/ui/carousel/carousel.ts
stevenle: See my note within carousel-synchronizer.ts. A way of conceptualizing it is if you imagine each carousel as an "island" and carousels that are synced are connected by a "bridge". If you need to pass a message to each connected island, visit each connected island and mark it as "visited". Then for every bridge on that island, check if you've already visited that island, and if not, visit that island.

If you search for "visit all linked nodes in undirected graph" you can find more examples with psuedocode.

src/ui/carousel/carousel-synchronizer.ts
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * Handle the transition of the given carousel to the given index.
+   *
+   * Will transition all other carousels synced with the given carousel to the
+   * given index.
+   *
+   * @param carousel
+   */
+  handleCarouselTransition(carousel: Carousel) {
+    const transitionTarget = carousel.getTransitionTarget();
```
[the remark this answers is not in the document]
angusm: Done, setup event data, removed TransitionTarget as a class, simplified to only store the slide HTML element locally and drivenBySync data is captured in the event data.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,536 @@
...
+    this.transitionToIndex(nextIndex);
+  }
+
+  /**
+   * Transition to the slide with the given index.
+   *
+   * @param index
+   * @param drivenBySync
+   */
+  transitionToIndex(index: number, drivenBySync: boolean = false): void {
```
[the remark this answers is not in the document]
angusm: Done.

src/ui/carousel/carousel.ts
stevenle: Sorry I misread your comment. Here's what I'm proposing:

```
a.sync(b)  // -> creates synchronizerA
a.carouselSynchronizer  // -> synchronizerA
b.caoruselSynchronzier // -> synchronizerA
```

```
c.sync(d)  // -> creates synchronizerB
c.carouselSynchronizer  // -> synchronizerB
d.caoruselSynchronzier // -> synchronizerB
```

When you do `c.sync(b)` I'd suggest one of two things: have some way of `synchronizerA` and `synchronizerB` merging its graph or simply throwing an error (which I mentioned in my original comment, "if both have a sync object, either find some way of merging the two objects so they're linked or take a lazy approach and just throw an error")

stevenle: FWIW I did mention in a previous comment that I felt it would be easier to just store a list of connected carousels within the carousel itself, in which case you wouldn't need to a way to manage the synchronizer objects.

src/ui/carousel/carousel-synchronizer.ts
angusm: Working demo pushed up. The issue I was seeing was a mistake I made while working on removing TransitionTarget, I wasn't passing a proper value through the data.

src/ui/carousel/carousel.ts
angusm: All good, I thought we had wanted to move away from merging sets, but I think I misunderstood the motivation of moving towards an undirected graph. Will write the merge code and then brush up on my knowledge of undirected graphs to fill in my gaps.

src/ui/carousel/carousel-synchronizer.ts
angusm: Removed CarouselSynchronizer. Syncing is now built into Carousel and related carousels stored as sets, merged as needed. All synced carousels share the same set instance.

src/ui/carousel/carousel.ts
angusm: Realized the rest of degu uses onRaf instead of loop. Renamed loop, which freed up the property name.

src/ui/carousel/carousel.ts
angusm: CarouselSynchronizer has been removed entirely. New setup within carousel handles this case.

src/ui/carousel/carousel.ts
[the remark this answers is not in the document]
angusm: Done. Added `goTo` as a convenience which can take either, and `goToSlide` and `goToIndex`.

angusm:
Hi Steven,

I think I've gotten all the changes together now. Thanks for bearing with me. Really like the renames, methods sound much simpler and clearer now.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,548 @@
...
+
+  /**
+   * Synchronize this carousel with the given carousels.
+   */
+  sync(...carousels: Carousel[]): void {
+    const masterSet = setf.merge(
+        this.syncedCarousels,
+        ...carousels.map((c) => c.syncedCarousels));
+    this.syncedCarousels = masterSet;
+    carousels.forEach((c) => c.syncedCarousels = masterSet);
```
stevenle, line 355: I think you need to do:

```
masterSet.forEach((c) => c.syncedCarousels = masterSet);
```

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,548 @@
...
+   * Go to the next slide.
+   */
+  next(): void {
+    this.transitionSlidesBy(1);
+  }
+
+  /**
+   * Go to the previous slide.
+   */
+  previous(): void {
```
stevenle, line 278: Can we rename this to `prev()`?

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,548 @@
...
+    this.autoplayTimeout = null;
+    this.syncedCarousels = new Set<Carousel>([this]);
+
+    this.init();
+  }
+
+  /**
+   * Returns the last known active slide.
+   */
+  getLastActiveSlide(): HTMLElement {
```
stevenle, line 155: What's the use case for this?

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,548 @@
...
+    return this.lastActiveSlide;
+  }
+
+  /**
+   * Change the active slide to the given slide by either index or element.
+   */
+  goTo(target: number|HTMLElement, drivenBySync = false) {
+    if (target instanceof HTMLElement) {
```
stevenle, line 163: Small nit: I would assume `if (typeof target === 'number')` is more performant for the interpreter but I could be wrong.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,548 @@
...
+    // requirements, but Carousel B didn't, A's slide 2X is equivalent to B's
+    // slide X, so if B transitions to X and A is already at 2X, we can safely
+    // ignore that sync, to prevent what will seem to the user like an
+    // unnecessary transition.
+    if (
+        this.getIndex(this.transitionTarget) % carousel.getSlideCount() !==
+        index
+    ) {
+      // If the current carousel has more slides than the syncing carousel,
+      // then we want to grab the closest equivalent slide.
```
stevenle, line 388: I had a bit of trouble following the logic in this particular block of code, I wonder if the code can be simplified somehow or if you could update the comment above to clarify with an example of what an "equivalent slide" means.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,548 @@
...
+        }
+      }
+
+      if (this.transitionTarget !== null) {
+        const hasTransitionedToTarget =
+            this.transition.hasTransitionedTo(this.transitionTarget);
+        if (hasTransitionedToTarget) {
+          this.transitionTarget = null;
+          this.eventManager.dispatch(
+              CarouselEvent.AFTER_CHANGE, this.generateEventData());
```
stevenle, line 499: I would suggest having different data objects for beforeChange and for afterChange, since in the "after" event, I would think the `currentSlide` is what it has already transitioned to and `nextSlide` is what it will be when the carousel transitions in X seconds.

(In the "after" event maybe it should only include `currentSlide`.)

Related: is it worth adding the index to this object? For most carousels that I've worked on in the past, knowing the current index number is probably more important than having the slide element.

stevenle (commented):
Thanks, Angus. This is feeling a lot better. A few small comments.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,548 @@
...
+}
+
+/**
+ * Options provided to the constructor
+ */
+export interface CarouselOptions {
+  activeCssClass?: string;
+  beforeCssClass?: string;
+  afterCssClass?: string;
+  allowLooping?: boolean;
```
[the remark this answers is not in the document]
angusm: Yes, done.

src/ui/carousel/carousel.ts
angusm: Good catch, thank you.

src/ui/carousel/carousel.ts
angusm: Done.

src/ui/carousel/carousel.ts
angusm: DraggableSlide calls it as a tie breaker when determining the active slide resulting from a drag.

If two slides are tied for distance as closest to the center, it will default to the one that was last active, not crucial, can remove.

src/ui/carousel/carousel.ts
angusm: You're right, confirmed with JS Bench https://jsbench.me/aokll2jqdm/1

src/ui/carousel/carousel.ts
angusm: Done.

src/ui/carousel/carousel.ts
angusm: Added more comments to clarify the motivation behind this code.

angusm:
Hi Steven,

Thanks for the feedback, most updates were straightforward, left the comment around the syncTo issue open, hopefully the additional comments clarify the code's purpose.

src/ui/carousel/carousel.ts
```diff
@@ -0,0 +1,614 @@
...
+  }
+
+  /**
+   * Sync this carousel to the given index in the given carousel.
+   * @param rawIndex
+   * @param carousel
+   * @private
+   */
+  private syncTo(rawIndex: number, carousel: Carousel) {
+    /**
```
stevenle, line 390: FYI The `/**` format for comments should be reserved for docstrings. You might be able to move this comment block into the main docstring since it's so long.

stevenle:
Thanks for the changes, the API for this looks good to me now. Only thing I might be concerned about for future-readiness is the ability to add/remove slides without having to dispose and re-initialize a carousel. I also wonder if we can potentially get away with assuming that the immediate children from a root element are the "slide" elements, and optionally provide an option for overriding that assumption so that the constructor could be as simple as `const c = new Carousel(el)`. But anyway if we leave it as is I'm fine with that.

I'm gonna mark +1 to this review, but I'll leave it to @uxder for the final approval to make sure everything fits with the current degu conventions.

stevenle (approved):

src/ui/carousel/carousel.ts
angusm: Done.

uxder (approved):
Thank you!  Let's check it in :)

PR #69, opened 2021-02-14
Create progress-watcher and inview-progress

uxder:
This PR has already been merged but adding it for TBR.

We currently write a lot of animations centered around progress but apart from raf-progress (which I would like to avoid going forward), don't have a good way setup callbacks for specific progress threshold (such as, I want to do X when the progress is between 0.2 - 0.4) and also lack a simple way to add css classes at specific progress points.

This PR adds two classes and updates css-parallax.

For clarity, I recommend viewing the files in the following order:

- ProgressWatcher
- InviewProgress
- CssParallax

You can mostly ignore the scrollToFocus updates.  That gets included because I'm comparing to main branch.

1) ProgressWatcher -  class to add callbacks for specific progress points.  Do X when progress is between 0.2 and 0.4.  or do X when progress cross 0.2.

2) InviewProgress - builds on ProgressWatcher.  I couldn't think of a better name so open to better names here. This is a class to add a css class to an element when the progress is between a range (such as 0.2 and 0.4).

3) Allow optional passing of inviewProgress in css-parallax to let css-parallax update the progress value of inviewProgress.

 Why?  It's really common to use css-parallaxer (which does css variable interpolations) with wanting to add css classes at specific trigger points.    Adding the option to pass inviewProgress to css-parallaxer makes setup easier without having to keep the progress between css-parallax and inviewProgress in sync.

This is an example of how you can use inviewProgress with cssParallax.
 ```ts
  const inviewProgress = new InviewProgress();
 
  // Add the class "active" when from range 0.2 - 0.4
  inviewProgress.add({
     range: [0.2, 0.4],
     element: el,
     className: "active"
  })
 
  // Setup css parallax
  const settings = {
     debug: false,
     top: '0px'
     bottom: '10px',
     inviewProgress: inviewProgress
 }
 const parallaxer = new CssParallaxer(el);
 parallaxer.init(settings, [...my css var interpolations])
 
 ```

### DIFF
--- a/src/dom/progress-watcher.ts
+++ b/src/dom/progress-watcher.ts
@@ -5,6 +5,7 @@ import { mathf } from '../mathf/mathf';
 export interface ProgressWatcherItem {
     range: number | Array<number>;
     callback: Function;
+    inactiveCallback?: Function;
     runWhen?: Function;
 }
 
@@ -21,6 +22,11 @@ export interface ProgressWatcherItem {
  *  this.progressWatcher.add({
  *    range: [0.2, 0.4],
  *    callback: (progress: number, direction: number)=> {
+ *          console.log(progress, direction);
+ *    }
+ *
+ *    // Runs when the progress is NOT 0.2 - 0.4
+ *    inactiveCallback: (progress: number, direction: number)=> {
  *          console.log(progress, direction);
  *    }
  *  })
@@ -131,6 +137,10 @@ export class ProgressWatcher {
                   watcher.callback(this.currentProgress, this.direction);
                 } else {
                   watcher.callback(this.currentProgress, this.direction);
+                }
+            } else {
+                if(watcher.inactiveCallback) {
+                  watcher.inactiveCallback(this.currentProgress, this.direction);
                 }
             }
         })
--- a/src/dom/progress-watcher.ts
+++ b/src/dom/progress-watcher.ts
@@ -1,0 +1,145 @@
+
+import { is } from '../is/is';
+import { mathf } from '../mathf/mathf';
+
+export interface ProgressWatcherItem {
+    range: number | Array<number>;
+    callback: Function;
+    runWhen?: Function;
+}
+
+
+
+/**
+ * A class that provides callbacks for progress points.
+ *
+ * ```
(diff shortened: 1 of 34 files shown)


src/dom/progress-watcher.ts
```diff
@@ -0,0 +1,165 @@
+
+import { is } from '../is/is';
+import { mathf } from '../mathf/mathf';
+
+export interface ProgressWatcherItem {
+    range: number | Array<number>;
```
stevenle, line 6: I think Google's TS style guide suggests using `number[]` for array types.

src/dom/progress-watcher.ts
```diff
@@ -0,0 +1,165 @@
...
+    callback: Function;
+    inactiveCallback?: Function;
+    runWhen?: Function;
+}
+
+
```
stevenle, line 12: Nit: There seems to be a lot of extraneous newlines in this file, and I noticed it's 4-space indented and I believe other files are using 2-spaces. Is there an automatic linter/fixer in this proj?

src/dom/progress-watcher.ts
```diff
@@ -0,0 +1,165 @@
...
+ *          console.log(progress, direction);
+ *    }
+ *  })
+ *
+ * // Triggers only when 0.2 is crossed in upwards or downwards direction.
+ *  this.progressWatcher.add({
+ *    range: 0.2,
+ *    callback: (progress: number, direction: number)=> {
+ *        // If going up
+ *        if(direction == -1) {
```
stevenle, line 39: Can you use an enum? https://www.typescriptlang.org/docs/handbook/enums.html

stevenle (commented):
I left a few small style nits. Didn't see any major issues with the API or the implementation.

src/dom/progress-watcher.ts
uxder: Done

src/dom/progress-watcher.ts
uxder: Done.

src/dom/progress-watcher.ts
uxder: I'll look into integrating google style rules for the project.

PR #77, opened 2021-04-29
Add callbacks to ScrollRenderFix

angusm:


uxder (commented):
Do you think we can update this pass a config object?

https://github.com/blinkkcode/degu/blob/main/src/ui/margin-outliner.ts#L91

I think it'll better future proof it in case we need to keep adding options.

uxder:
LGTM!  Thank you!

PR #78, opened 2021-05-05
Add a script loader

stevenle:
The script loader is meant to be used for 3rd party scripts that should be loaded upon interaction, such as the YouTube API. Example usage is something like:

```javascript
const scriptLoader = new ScriptLoader();
scriptLoader.load('https://www.youtube.com/iframe_api', {
  test: () => window.YT,
}).then(() => {
  // do something with the youtube api
});
```

With the example code above, if the `test` function returns `true` right from the get-go, the script isn't added to the page at all. A class is used here so that the `renderDom()` method can be overridden for special use cases, like if a site is enabled for CORS and needs to be safe-loaded through some special mechanism.

A similar utility can be found in the Closure Library here: https://github.com/google/closure-library/blob/master/closure/goog/net/jsloader.js

Let me know if any feedback, happy to address.

### DIFF
--- a/src/loader/script-loader.ts
+++ b/src/loader/script-loader.ts
@@ -30,6 +30,7 @@ export interface ScriptLoaderConfig {
  */
 export class ScriptLoader {
   private loadedScripts: Record<string, Promise<void>> = {};
+  private disposed = false;
 
   /**
    * Loads a script element onto the page.
@@ -59,6 +60,10 @@ export class ScriptLoader {
       const startTime = time.now();
       const timeout = options.timeout || 5000;
       const callback = () => {
+        if (this.disposed) {
+          reject('script loader is disposed');
+          return;
+        }
         if (options.test()) {
           resolve();
           return;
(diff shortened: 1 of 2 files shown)


src/loader/script-loader.ts
```diff
@@ -0,0 +1,99 @@
...
+    return promise;
+  }
+
+  /**
+   * Renders the DOM for the script element. Subclasses should override this
+   * method for more complex script renderings.
+   */
+  protected renderDom(url: string): void {
+    const script = document.createElement('script');
+    script.src = url;
```
uxder, line 91: Isn't there a script on load event we can use instead of Raf?

uxder (commented):
Looks good.  Just some small question / comments.

src/loader/script-loader.ts
uxder: https://github.com/google/closure-library/blob/master/closure/goog/net/jsloader.js#L186

stevenle: There is but I need to keep this agnostic of that, reason being that on gstore we only have access to inject scripts via the `mannequinService.safeLoad()` method. Also the onload isn't particularly deterministic; there's a note in jsloader that says something along the lines of "the caller needs to ensure the script is actually successfully loaded".

uxder: ah okay gotcha.

uxder:
LGTM :)

PR #90, opened 2021-05-18
test: add eslint to test workflow

stevenle:


stevenle:
@uxder I think this is failing from the coverage line. Mind taking a look? I also noticed the package.json has a "module" declaration although the build doesn't actually output an esm file.

uxder:
Okay issue should be fixed now.

PR #102, opened 2021-05-26
chore: cleanup index exports

Zoramite:


jeremydw:
This looks good as it's a non-breaking change and doesn't result in any usage changes, right?

Also (this comment is external to this change), but as a practice I think we should export everything that might be used by devs using degu. Then we can clean up our package/usage paths and remove `lib` from the import paths in projects.

In Amagaki I copied what @Zoramite did and package import paths are now fixed:
https://github.com/blinkk/amagaki/blob/main/src/index.ts

I also added a basic test to verify this but it may not be necessary for degu: 
https://github.com/blinkk/amagaki/blob/main/tests/packagePath/amagaki.ts

Zoramite:
Yes, this should be a non-breaking change. It just 'simplifies' how the export happens so that the import and export do not have to be separate statements.

PR #125, opened 2021-07-22
feat: Start subsequent raf run if necessary

angusm:
This allows for an earlier step to be scheduled during
the callback for a later step. For example scheduling a
read() callback during a postWrite() callback.


uxder (approved):
LGTM

PR #161, opened 2021-10-27
feat: Carousel pauses if condition not met

angusm:
Carousels with conditions set by inview were immediately
transitioning to the next slide as soon as they were visible.
This was being caused by the autoplayTimeout continuing to
tick down while the carousel was out of view.


micjamking:
@angusm LGTM! Thanks for getting to this!

micjamking (approved):

PR #175, opened 2021-11-18
Add degu-image component

uxder:
Adds the following changes:
- add the first lit component and testing setup.  We have existing ava tests so `npm run test` runs all test. `npm run test-lit` runs only lit component tests
- Adds `<degu-image>` component
- Adds initial test that covers autowidth
- Additional tests that covers non-autowidth images will be in a follow up PR.

Questions:
- there are basically two modes - autowidth (FIFE images) and picture - srcset (non-autowidth).  Do we need a third mode that just renders an image with loading="lazy"?
- let me know if there are any features missing


uxder:
@jeremydw @stevenle - bringing this back up.

jeremydw:
Core implementation looks good and I like the idea of having three modes:
- Normal image tag
- Google "Image Service" images with autowidth
- Picture tag with srcset (don't see us really practically using this, though -- so it's OK to omit it)

Two more nits:
1. Change "FIFE" to "Google Image Service"
2. @stevenle @uxder thoughts on changing `a11y-label` back to `alt` for the generic implementation? The reason I bring this up is because the Next Image tag (https://nextjs.org/docs/api-reference/next/image) conceptually is an "enhanced normal image tag" -- e.g. all of the builtin attributes work as expected without the user needing to remember to convert "alt" to "a11y-label".

uxder:
> 2. @stevenle @uxder thoughts on changing `a11y-label` back to `alt` for the generic implementation? The reason I bring this up is because the Next Image tag (https://nextjs.org/docs/api-reference/next/image) conceptually is an "enhanced normal image tag" -- e.g. all of the builtin attributes work as expected without the user needing to remember to convert "alt" to "a11y-label".

This sounds good to me.

uxder:
Okay I've updated to Google Image Service and also changed a11y-label over to `alt`

src/lit/image.ts
```diff
@@ -110,7 +110,8 @@ export class DeguImage extends LitElement {
     this.isFife =
```
jeremydw, line 110: Update this varname too, e.g. `this.isGoogleImage`

uxder: Oops.. Thank you.  Done

jeremydw (approved):

PR #176, opened 2021-11-24
Add timer 

uxder:
Changes:

- Add a timer class supporting subscription to set duration events, pause and unpausing
- Update events to accept Function type.  EventCallback is an unnecessary abstraction that isn't adding anything


uxder:
Merging this as TBR.  Let me know if there are any concerns.

PR #180, opened 2021-11-25
Add lit degu-video component

uxder:
Changes:
- Add lit degu-video component
- Move all degu lit component example / sample pages to one page


src/lit/video.ts
```diff
@@ -0,0 +1,209 @@
...
+ *       width: 100%;
+ *   }
+ * ```
+ */
+export class DeguVideo extends LitElement {
+  @property({type: String, attribute: 'src'})
+  src: string;
+
+  @property({type: String, attribute: 'a11y-label'})
+  private a11yLabel: string;
```
jeremydw, line 53: Let's call this `aria-label` to be consistent with "extending the default element"?

src/lit/video.test.ts
```diff
@@ -0,0 +1,64 @@
...
+import * as func from '../func/func';
+
+// Define it.
+!window.customElements.get('degu-video') &&
+  window.customElements.define('degu-video', DeguVideo);
+
+const addStylesToPage = func.runOnlyOnce(() => {
+  dom.addStylesToPage(`
+    degu-video {
+      display: inline-block
```
jeremydw, line 13: Small nit but I think this is missing semicolons at the end, and fix the inconsistent indentation

examples/index.html
```diff
@@ -9,7 +9,7 @@
   <h1>Web Components</h1>
 
   <ul>
-    <li><a href="/degu-image.html">Degu Image</a>
+    <li><a href="/lit-components.html">Lit Components</a>
```
jeremydw, line 12: Make the example filenames consistent with the actual filenames, so maybe make this `lit.html` to be consistent with `src/lit` ?

examples/lit-components.html
```diff
@@ -0,0 +1,70 @@
...
+  <div class="section">
+    <div class="section__header">Degu Image</div>
+    <degu-image src="https://placeholder-dot-madebygoog.appspot.com/2650x720.svg" alt="Logo" class="header__logo__image"></degu-image>
+    <degu-image
+      width="2650" height="720"
+      src="https://placeholder-dot-madebygoog.appspot.com/2650x720.svg" alt="Logo" class="header__logo__image"></degu-image>
+
+    <degu-image
+      width="2300" height="958"
+      src="https://lh3.googleusercontent.com/EtrsUA5VUfcBjJZqcgfXBLQfvE9lkYMnaU9L8Zp_Y0MzIciD3NDXhYkakb_59Bk8oSCHMfpsDXGJtzrtY9aDYZT3xFic11kBgwM" alt="Logo" class="header__logo__image"></degu-image>
```
jeremydw, line 52: Use placeholder assets not tied to a specific client project

src/lit/video.test.ts
uxder: Done

examples/lit-components.html
uxder: Done

examples/index.html
uxder: Done

src/lit/video.ts
uxder: Some concerns with this change.

So right now, it renders as follows

```html
<degu-video a11y-label="Video Aria Label" >
      <video role="img" aria-label="Video Aria Label" load="true">
         ...
      </video>
    </degu-video>
```

This gets announced as "Video Aria Label : Image" (given the role="img").

The change will modify this.  We end up with two aria labels, which gets announced as a single "Video Aria Label: Group" (in VO) - VO doesn't seem to focus into the video.

```html
<degu-video aria-label="Video Aria Label" >
      <video role="img" aria-label="Video Aria Label" load="true">
         ...
      </video>
    </degu-video>
```

We could "remove" the top level aria-label from degu-video but this would also have to make it a non-reactive property.

jeremydw: Ah right, sorry, I forgot about this concern. OK - another alternative. What if we just added `role=img` to the host element? Then `aria-label` could be left unmodified.

stevenle: FWIW I think we should come up with some convention that will work across all custom elements while preserving as much of the native experience as we can.

Downside of applying `role=img` to the custom element is that we'll probably need to do something like `` to avoid duplicate reading.

Need to also consider something like `<custom-button>` which would have an embedded button, we shouldn't really add `role="button"` to that one.

jeremydw: Video is a bit unique because the builtin element doesn't have consistent screen reader support, so the approach we want to take with it is:

```<video role="img" aria-label="Aria label">```

I think if we approach building our custom elements as "enhancing the functionality of builtin elements" we should try to stay aligned with the built-in attributes, roles, etc. -- so I think it'll have to be done on a case-by-case basis depending on the element?

For video specifically I think when we checked this last, the videos weren't announced. But yes, we'd need to add `<video aria-hidden="true">` to enforce that.

uxder: Okay updated and also added some test.  I tested it on VO and it's working fine.

jeremydw (approved):

uxder:
Released via: 1.27.0

PR #184, opened 2021-11-28
Add lit component boostrap

uxder:
Adds a simple way to declare all degu lit components.

```
 import { register} from '../lib/lit/lit';
 
 // Called once.
 register();
 ```
 
 Drawbacks:
 - imports all degu lit components which could be large as the library grows

### DIFF
--- a/examples/degu-image.js
+++ b/examples/degu-image.js
@@ -1,9 +1,9 @@
 
- import { DeguImage} from '../lib/lit/image';
+ import { register} from '../lib/lit/lit';
 
 export default class DeguImageSample {
 
   constructor() {
-    window.customElements.define('degu-image', DeguImage);
+    register();
   }
 }
--- a/src/lit/lit.ts
+++ b/src/lit/lit.ts
@@ -1,0 +1,12 @@
+import {DeguImage} from './image';
+
+const defaultComponents: Record<string, typeof HTMLElement> = {
+  'degu-image': DeguImage,
+};
+
+export const register = (components = defaultComponents) => {
+  for (const key in components) {
+    !window.customElements.get(key) &&
+      window.customElements.define(key, components[key]);
+  }
+};

### CODE REVIEW

File: examples/degu-image.js [hunk 1]
```diff
@@ -1,9 +1,9 @@
 
- import { DeguImage} from '../lib/lit/image';
+ import { register} from '../lib/lit/lit';
 
 export default class DeguImageSample {
 
   constructor() {
-    window.customElements.define('degu-image', DeguImage);
+    register();
```
jeremydw, line 7: I'm not seeing the value to this at this time (but may after the library grows?). I think users may still have some value in enumerating exactly which components they want to use in their projects (versus bringing all of them in).

One nit here is I would name this `defineAll` (or similar) so it's clearer what it's doing.

Thoughts @stevenle ?

stevenle: I do like the idea of being able to pick and choose which components since a project may only use a couple, especially if we eventually have very complex components that are more project-specific (like a degu-lottie-player components which not a lot of sites will use).

I'm fine with `defineAll` but I would suggest having something like "Components" (e.g. `defineAllComponents` or `defineComponents`) in the name of the method since it helps readability when it's imported and used in another file.

stevenle:
Sorry I'm commenting on this a little late (had a big, short-deadline PGO project last week), but any chance we could avoid using `lit` in the project structure? I feel like web components can and generally will live on in degu's life, but we may potentially move away from lit eventually and so I feel like it's better for us to try to be implementation-agnostic in our naming if we can.

examples/degu-image.js
stevenle: @jeremydw Related to the comments above, I think the plugin system you're creating with ama should have some way of defining its degu component dependencies so that a caller of the plugin will just need to do something like the following instead of having to read some README to figure out which degu components are required dependnecies.

```typescript
import {FooPlugin} from './ama/plugins/foo';
FooPlugin.defineAllComponents();
```

uxder:
Yeah we can move everything to maybe a `components` folder?  

I'll do that in a follow up PR later after we have closed out the `degu-video` and `degu-youtube-modal` PRs.

jeremydw:
> Yeah we can move everything to maybe a `components` folder?
> 
> I'll do that in a follow up PR later after we have closed out the `degu-video` and `degu-youtube-modal` PRs.

Either "components" or "elements" sounds good to me. :)

examples/degu-image.js
jeremydw: I can figure this out with the plugin; but note that even in the plugin we wouldn't want to define all components since only some may be used. I'll defer to working on this in the plugin until if/when we move it to a package

PR #190, opened 2021-12-08
feat: Add Ui swipe

uxder:
Adds the following:

1) ui/swipe - used to easily monitor left/right/up/down swipe on a given element

2) add swipe events to dom-watcher

### DIFF
--- a/src/ui/swipe.ts
+++ b/src/ui/swipe.ts
@@ -1,0 +1,164 @@
+import {DomWatcher} from '../dom/dom-watcher';
+import {EventCallback, EventManager, EventDispatcher} from './events';
+import * as dom from '../dom/dom';
+
+export enum SwipeEvent {
+  UP = 'swipeUp',
+  DOWN = 'swipeDown',
+  LEFT = 'swipeLeft',
+  RIGHT = 'swipeRight',
+}
+
+/**
+ * A basic class that allows you to watch directional swipe events on a given
+ * element.
+ *
+ *
+ * Usage:
+ * ```ts
+ * const swiper = new Swipe(myElement);
+ *
+ * swiper.on(Swipe.Events.UP, ()=> {
+ *   // swipe up.
+ * })
+ * swiper.on(Swipe.Events.LEFT, ()=> {
+ *   // swipe left
+ * })
+ *
+ * // Later
+ * swiper.dispose();
+ * ```
+ *
+ *
+ *
+ * Events are also fired on the element itself.
+ *
+ * ```ts
+ * const swiper = new Swipe(myElement);
+ * myElement.addEventListener(Swipe.Events.UP, ()=> {
+ *   // swipe up.
+ * })
+ * ```
+ */
+export class Swipe implements EventDispatcher {
+  private rootElement: HTMLElement;
+  private eventManager: EventManager;
+  private watcher: DomWatcher;
+  private x: Number;
+  private y: Number;
+  private isSwiping = false;
+  private allowSwipe = false;
+  private debounceDuration = 300;
+
+  constructor(rootElement: HTMLElement) {
+    this.rootElement = rootElement;
+    this.eventManager = new EventManager();
+    this.watcher = new DomWatcher();
+
+    const isTouchSupported = 'ontouchstart' in document.documentElement;
+
+    // Add listeners
+    if (isTouchSupported) {
+      this.watcher.add({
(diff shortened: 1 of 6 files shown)


stevenle:
An initial thought from looking at the code structure is that anyone who imports the DomWatcher will now import the entire swipe/gesture library, which may not be needed for majority of use cases. I'd suggest decoupling the DomWatcher from the rest of the code if possible (or you could subclass DomWatcher and apply this functionality on top if you prefer the current direction).

uxder:
Yeah that's a good point.  Since it's probably not that common to need swipe events, I think you are right in that we don't need it bundled with Dom-watcher.  I've gone and removed it.

stevenle:
Thanks, Scott. I think if we do want to bundle these gesture recognizers together (similar to HammerJS), we can probably create something like `GestureWatcher` which extends `DomWatcher`.

Will check the code for style and correctness shortly.

src/ui/swipe.ts
```diff
@@ -0,0 +1,164 @@
...
+ * myElement.addEventListener(Swipe.Events.UP, ()=> {
+ *   // swipe up.
+ * })
+ * ```
+ */
+export class Swipe implements EventDispatcher {
+  private rootElement: HTMLElement;
+  private eventManager: EventManager;
+  private watcher: DomWatcher;
+  private x: Number;
```
stevenle, line 47: `Number` => `number`

src/ui/swipe.ts
```diff
@@ -0,0 +1,164 @@
...
+        eventOptions: {passive: true},
+        callback: this.onPointerMove.bind(this),
+      });
+    }
+  }
+
+  static Events = SwipeEvent;
+
+  private onPointerDown(e: TouchEvent) {
+    this.x = (e.touches && e.touches[0].clientX) || e['x'];
```
stevenle, line 105: Are you using `e['x']` because `x` isn't defined on `TouchEvent`? I wonder if you should declare `(e: TouchEvent | MouseEvent)` above in that case.

src/ui/swipe.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (!this.x || !this.y || this.isSwiping || !this.allowSwipe) {
+      return;
+    }
+
+    this.isSwiping = true;
+
+    const currentX = (e.touches && e.touches[0].clientX) || e['x'];
+    const currentY = (e.touches && e.touches[0].clientY) || e['y'];
+
+    const diffX = currentX - +this.x;
```
stevenle, line 124: Per style guide, use `Number(this.x)` instead.

(Although looking at this more closely, why is the number conversion here even necessary? In what cases would this not be a number?)

src/ui/swipe.ts
uxder: Done

src/ui/swipe.ts
uxder: Removed.  (it's not needed - I ported this and missed that)

src/ui/swipe.ts
```diff
@@ -0,0 +1,164 @@
...
+      } else {
+        this.eventManager.dispatch(Swipe.Events.DOWN);
+        dom.event(this.rootElement, Swipe.Events.DOWN, {});
+      }
+    }
+
+    window.setTimeout(() => {
+      this.isSwiping = false;
+      this.allowSwipe = false;
+    }, this.debounceDuration);
```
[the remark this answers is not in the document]
uxder: Good call out.  It's actually not needed at all so I've removed it.

src/ui/swipe.ts
```diff
@@ -0,0 +1,163 @@
...
+    this.allowSwipe = false;
+  }
+
+  private onPointerMove(e: TouchEvent) {
+    if (!this.x || !this.y || !this.allowSwipe) {
+      return;
+    }
+
+    const currentX = (e.touches && e.touches[0].clientX) || e['x'];
+    const currentY = (e.touches && e.touches[0].clientY) || e['y'];
```
stevenle, line 124: Update to `e.x` and `e.y`.

stevenle (approved):

src/ui/swipe.ts
uxder: Done - Had to check if it was a instance of to avoid typescript errors.

src/ui/swipe.ts
uxder: Done

PR #198, opened 2022-01-12
fix: degu-image should calculate the max of the container and image

uxder:
Degu image can miscalculate autowidth sizes if the developer has css that makes the image larger than the `degu-image` container.

For example:

```
<degu-image> <img></degu-image>

degu-image img
  position: absolute
  width: 100%
  height: 100%
```
Results in the container offsetWidth being smaller than the image itself.

This fixes problem by taking the larger of the container and image as the basis of autowidth calculation.


uxder:
Going to TBR this.

Please note any issues / concerns.  Thanks

PR #223, opened 2022-03-15
feat: Add tabindex

uxder:
Feature to add tabindex based on preset condition (such as breakpoints).

This is a similar concept to atClass but applies to tabindex.

```html
  <div at-tab-index="2@tablet-lt 3@laptop-gt">
     Tab Index gets set 2 on tablet-lt and 3 on laptop-gt
  </div>
 
  <div at-tab-index="2@mobile">Tab index only on mobile</div>
```

### DIFF
--- a/examples/at-tabindex.html
+++ b/examples/at-tabindex.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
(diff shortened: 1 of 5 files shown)


uxder:
This is PR is a TBR.  Let me know if you have any concerns.

Thanks

PR #235, opened 2022-04-27
Add support for mediaMatch in DomWatcher

angusm:
Hi Scott,

A change that adds some extra typing but mainly adds support for passing the result of a `window.matchMedia()` call to DomWatcher to listen for events.


uxder (commented):
Ah I didn't think about this.  We can just listen to the change event with this.  Nice.

I think this is nice but can you add some quick examples to the doc.

uxder (approved):
LGTM. ty!  

I think you'll need to squash merge this with the release please style commit messages to auto release it.

PR #256, opened 2022-07-14
Support checks on slot in closestElementComposed

angusm:
Check for closest while factoring in the assigned slot.

Handles a case like:
```
<custom-element>
  <tile slot="tile"></tile>
</custom-element>
```

Where `<custom-element>` renders:
```
render() {
  return html`
    <mosaic radius="20">
      <slot name="tile"></slot>
    </mosaic>
  `;
}
``

Where the `tile` is looking to pull its radius from a parent `mosaic` element.

### DIFF
--- a/src/dom/dom.ts
+++ b/src/dom/dom.ts
@@ -1221,7 +1221,14 @@ export function closestElementComposed(selector, baseElement) {
     if (!baseElement || baseElement === document || baseElement === window)
       return null;
     const found = baseElement.closest(selector);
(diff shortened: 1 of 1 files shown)


uxder (approved):
LGTM.

angusm:
Does not solve the original problem as `assignedSlot` is not set when `render()` is run. This may not help address the issue of parent/child communication through ShadowDOM. Going to investigate use of ContextProtocol instaed.

PR #271, opened 2022-10-25
fix: make youtube api publicly accessible

hellomichael:
Hey @uxder,

Have a use case where I need to jump to specific timestamps with [seekTo](https://developers.google.com/youtube/iframe_api_reference#seekTo). Can we make the Youtube Api publicly accessible for this component?


uxder:
LGTM

PR #283, opened 2023-01-19
release!: v3.0.0

stevenle:
This PR is a set of breaking changes for a `v3.0.0` release of degu.

Updates include:

* Switched from `tsc` to `tsup` for compiling/bundling degu files into `lib/` for distribution on npm
* Updated imports to be more compatible with node16 style imports (which require the `.js` extension)
* Removed a few deprecated functionality and obsolete functions
* Removed the ES5-style namespace exports and switched to wildcard imports of top-level functions instead (which should help with tree-shaking)
* Added a MIGRATIONS.md file for info on how to upgrade from v2.x.y to v3.x.y
* Moved 3rd party deps (three, pixi, angular, lit, lottie-web, etc.) to optional `peerDependencies`


uxder:
Do you know if the change from `import {DomWatcher} from '@blinkk/degu/lib/dom/dom-watcher'` to `import {DomWatcher} from '@blinkk/degu'` would impact file size?    I know in theory it shouldn't but in our larger projects, I recall there were some treeshaking issues importing from index (which might be a factor of esbuild)

I think from a developer perspective, importing from index is much easier although we'll need to a wide search and replace to migrate things on our larger projects.  We'll also probably need add stuff to `src/index.ts` because many classes are missing from that file.

uxder (approved):
Cool - thanks for making these changes.  LGTM!

stevenle:
Updated PR title/description, this one's ready for another review @uxder !

uxder (approved):
LGTM.  Thank you for putting this together!

PR #292, opened 2023-01-27
refactor: switch from webpack to vite for examples

stevenle:
This is a major cleanup that removes a few legacy dependencies (webpack, gts, etc.) and migrates to modern libs that are more performant and well-maintained (vite).

Note: As part of this cleanup, I've temporarily disabled eslint, will create a follow-up PR that has a new eslint config that can be reviewed.

@uxder please review the changes in package.json to make sure I didn't remove anything by mistake.


stevenle:
Note: after this change, you might want to refactor the examples so that they don't rely on the monolith at index.js and instead each example should individually import its own js file. Also with the vite change, you would be able to do `<script type="module" src="./foo.ts"></script>` within `examples/foo.html` and vite should automatically transpile the file correctly. You can then import files directly from `src/` (instead of `lib/`) which enables proper type checking and doesn't require you to re-run the build while you're in development.

uxder (approved):
LGTM - didn't see any issues with package.json.

TY!

PR #297, opened 2023-02-01
fix: fix playback error in lazy-video

uxder:


src/dom/lazy-video.ts
```diff
@@ -242,7 +242,9 @@ export class LazyVideo {
   private playVideo() {
     if (!dom.testVideoIsPlaying(this.video)) {
       const playPromise = this.video.play();
-      playPromise.then(() => {}).catch();
+      if (playPromise !== undefined) {
```
stevenle, line 245: I think you can shorten this a little to something like:

```
if (playPromise) {
  playPromise.catch(() => {});
}
```

stevenle (approved):

src/dom/lazy-video.ts
uxder: Updated and good point.     I was kind of blindly following the notes here: https://developer.chrome.com/blog/play-request-was-interrupted/

PR #300, opened 2023-04-22
feat: add "orientation" option to youtube modal

stevenle:
Some sites are now embedding "YouTube Shorts", which are a portrait orientation. A user would be able to update the yt player to use portrait by setting `data-youtube-modal-orientation="portrait"` to the trigger element.

### DIFF
--- a/src/ui/youtube-modal/youtube-modal.ts
+++ b/src/ui/youtube-modal/youtube-modal.ts
@@ -71,6 +71,11 @@ interface PlayOptions {
    * The time in seconds to offset the video when the video plays.
    */
   startTime?: number;
+
+  /**
+   * YT player orientation. Defaults to "landscape".
+   */
+  orientation?: string;
 }
 
 // Default configuration.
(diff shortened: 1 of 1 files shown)


src/ui/youtube-modal/youtube-modal.ts
```diff
@@ -71,6 +71,11 @@ interface PlayOptions {
    * The time in seconds to offset the video when the video plays.
    */
   startTime?: number;
+
+  /**
+   * YT player orientation. Defaults to "landscape".
+   */
+  orientation?: string;
```
jeremydw, line 78: Is this just for me to style with CSS when I call it; or does it impact the default player somehow? I traced through `playOptions` usage and didn't see it referenced anywhere.

stevenle: Yeah just for CSS. The YT iframe player will automatically adapt to its orientation.

jeremydw: 👍 Might be good to clarify in the comment that it just sets the data attribute on the player (rather than impacting anything else). Other than that nit lgtm

jeremydw (approved):

